PR #3, opened 2024-07-23
Soup to Svg function

imrishabh18:
@seveibar Quick question: To generate the svg from soup before going to png, I was going the direction of writing all the svg functions to render component. I think this is the wrong way right? It must be generated from the `schematic-viewer` and `pcb-viewer` right?

But if that is the case then, why is this package named `soup-to-svg`?


seveibar:
@imrishabh18 you're basically doing it correctly, we shouldn't use schematic-viewer etc. this is an independent package

imrishabh18:
Ohh, then we need to use the `routing` package here as well for the traces?

seveibar:
No the traces are pre-routed via the soup

imrishabh18:
@seveibar This is for this [soup](https://debug.tscircuit.com/soup_group/builder:%20(rishabh)%20no%20traces%20through%20net%20labels#%7B%22selected_engine%22%3A%22schematic%22%2C%22selected_layout_index%22%3A0%7D). I think this can be good point to merge as the main schematic components are added and rendered as svg.

This will need a lot of work, which could be complex(lots of hit and try needed) it seems on fine tuning the positions in svg as we add more components.

seveibar:
Yes, in the next PR let's add storybook so that it's easy to do visual regression testing and have a viewer.

seveibar (approved):

seveibar:
also a format script, typescript ci, format ci etc. (available with @tscircuit/plop) just to make sure it's easy to contribute

The PCB render is more important than the schematic render, i would focus most of your efforts on PCB rendering. You'll need to have two modes for outputing, one for schematic and one for PCB.

Also at some point we should try to horizontally scale this codebase- meaning we should try to have small files/functions for different operations. Aider/Claude will by default create large files, but these are bad for collaboration (eventually)

imrishabh18:
Sure, I will make those changes and do a refactor

PR #5, opened 2024-07-24
Storybook added

imrishabh18:
In the storybook stories, I have set the svg directly to the html. Should we switch to the `Schematic` viewer and pass the soup or this is fine?
Screenshot 2024-07-24 at 10 27 32 AM


src/stories/net-label-not-overlap.stories.tsx
```diff
@@ -500,8 +501,17 @@ const soup = [
...
         source: null,
     },
-];
+  ];
+  const svg = soupToSvg(soup);
 
-const svg = soupToSvg(soup);
+  return (
+    <div
+      dangerouslySetInnerHTML={{ __html: svg }}
```
seveibar, line 509: this is 100% correct, nice

seveibar (approved):

PR #7, opened 2024-07-24
feat: svg to png function added

imrishabh18:
Closes - https://github.com/tscircuit/tscircuit/issues/261


imrishabh18:
After all the pieces are stitched up, and the end to end flow is fine. Will try to refactor this repo, so it's for other packages to use this

seveibar (changes requested):
we don't want to build in png compatibility to this library, we can create `@tscircuit/soup-to-png` instead.

Also we shouldn't use `svgexport`, it's not a popular npm library and is nodejs-only (so breaks browser compatibility for this module), and we generally should avoid any shell calls in our libraries

seveibar:
Alternative libraries with a better interface and support:

* https://www.npmjs.com/package/@resvg/resvg-wasm
* https://www.npmjs.com/package/convert-svg-to-png

imrishabh18:
Makes sense, will create another repo `@tscircuit/soup-to-png` then.

PR #12, opened 2024-08-10
Fix/schematic port

imrishabh18:
Closes - https://github.com/tscircuit/circuit-to-png/issues/12

- Adds the `schematic_port` and parse the text appropriately.
- Flips the schematic in Y-axis to match the schematic-viewer

Screenshot 2024-08-11 at 2 20 50 AM

### DIFF
--- a/src/lib/soup-to-svg.ts
+++ b/src/lib/soup-to-svg.ts
@@ -1,74 +1,80 @@
 import type { AnySoupElement } from "@tscircuit/soup";
 
 function soupToSvg(soup: AnySoupElement[]): string {
-  const svgContent: string[] = [];
   let minX = Number.POSITIVE_INFINITY;
   let minY = Number.POSITIVE_INFINITY;
   let maxX = Number.NEGATIVE_INFINITY;
   let maxY = Number.NEGATIVE_INFINITY;
 
-  // Process components
-  for (const component of soup.filter(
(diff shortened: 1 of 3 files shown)


imrishabh18:
@seveibar For the last trace edge to connect to the port, I was debugging but not sure what does the `ti` in these array mean. If I am not ignoring the `ti` edge then I am seeing something like this, where the shortest trace is getting added. 

Screenshot 2024-08-11 at 2 22 04 AM
Screenshot 2024-08-11 at 2 23 27 AM

seveibar:
@imrishabh18 I'm not sure what ti means (it might be an abbreviation for "... index"), but I think it's out-of-spec so you can probably ignore it.

As for the line, it seems to me like you're drawing the line path such that it's closing itself, the schematic route paths should never self-close. They may need to connect the first/last segment to the x/y of the port, that I'm not sure of. Looking at the spec to see if there's any details...

seveibar:
Yea my guess is if the trace isn't touching the port you basically are supposed to manually connect it by searching for the relevant `schematic_port`'s x/y using `from_schematic_port_id` or `to_schematic_port_id`

image

imrishabh18:
Fixed the last trace edge issue @seveibar 

Screenshot 2024-08-11 at 6 40 58 PM

seveibar (approved):

Issue #18, opened 2024-08-21
Switch to using svgson for better types


seveibar:
https://github.com/elrumordelaluz/svgson

PR #21, opened 2024-08-21
feat: svgson added

imrishabh18:
Closes - #18

### DIFF
--- a/src/lib/pcb-soup-to-svg.ts
+++ b/src/lib/pcb-soup-to-svg.ts
@@ -1,6 +1,14 @@
 import type { AnySoupElement } from "@tscircuit/soup";
+import { stringify, type INode } from "svgson";
 import { applyToPoint, compose, scale, translate } from "transformation-matrix";
-import { stringify } from "svgson";
+
+interface SvgObject {
+  name: string;
+  type: 'element' | 'text';
+  attributes?: { [key: string]: string };
+  children?: SvgObject[];
+  value?: string;
+}
 
 function pcbSoupToSvg(soup: AnySoupElement[]): string {
   let minX = Number.POSITIVE_INFINITY;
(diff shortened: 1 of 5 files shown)


src/lib/pcb-soup-to-svg.ts
```diff
@@ -85,53 +118,127 @@ function pcbSoupToSvg(soup: AnySoupElement[]): string {
...
 
-function createPcbComponent(component: any, transform: any): string {
+function createSvgElement(item: AnySoupElement, transform: any): any {
+  switch (item.type) {
+    case 'pcb_component':
+      return createPcbComponent(item, transform);
+    case 'pcb_trace':
+      return createPcbTrace(item, transform);
+    case 'pcb_plated_hole':
+    case 'pcb_smtpad':
```
seveibar, line 128: i forgot to mention this but smtpads aren't holes so this is a bit weird

src/lib/soup-to-svg.ts
```diff
@@ -40,56 +41,75 @@ function soupToSvg(soup: AnySoupElement[]): string {
...
+        type: 'element',
+        children: [
+          {
+            type: 'text',
+            value: `
+              .component { fill: none; stroke: red; stroke-width: 0.03; }
+              .component-pin { fill: none; stroke: red; stroke-width: 0.03; }
+              .trace { stroke: green; stroke-width: 0.03; fill: none; }
+              .text { font-family: Arial, sans-serif; font-size: 0.2px; }
+              .port { fill: none; stroke: blue; stroke-width: 0.03; }
```
seveibar, line 103: i think we should not use `<style />` and just put the properties directly on the svg objects, because theoretically this stuff will change for different layers, but issue for another day, just making a note

seveibar (approved):
hell yea

src/lib/pcb-soup-to-svg.ts
imrishabh18: It was getting handled here, https://github.com/tscircuit/circuit-to-svg/blob/feat/add-svgson/src/lib/pcb-soup-to-svg.ts#L175

But I will make a different function itself

imrishabh18:
Good to merge? @seveibar

seveibar:
Yep!

PR #25, opened 2024-08-26
feat: Integrate schematic symbols

imrishabh18:
What do you think about this of using the `circuitJsonToSchematicSvg` as a Promise, cause the schematic-symbols is returned as a string and need to be `parse` by svgson?

Also, currently the [rotateSymbol](https://github.com/tscircuit/schematic-symbols/blob/43ef01b0161609f4ec792a95363368f99a5f405e/drawing/rotateSymbol.ts#L20) is not exported as well as not rotating the symbols by the value you pass. So, I think this needs to be an issue to work on.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,16 @@
 {
   "name": "circuit-to-svg",
-  "version": "0.0.3",
+  "version": "0.0.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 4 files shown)


seveibar:
@imrishabh18 why not use `parseSync`?

imrishabh18:
Updated it, didn't check if `parseSync` was available

seveibar (approved):
Woot!! Going to try this with core today

PR #28, opened 2024-08-29
Fix/schematic symbols attributes

imrishabh18:
Screenshot 2024-08-30 at 2 31 26 AM

Attributes which are working for the symbols,
- position
- width, height
- rotation

### DIFF
--- a/src/lib/soup-to-svg.ts
+++ b/src/lib/soup-to-svg.ts
@@ -147,12 +147,57 @@ function createSchematicComponent(
 ): any {
   const transform = `translate(${center.x}, ${center.y}) rotate(${(rotation * 180) / Math.PI})`;
 
-  if(symbolName) {
-    // TODO: Change the types in soup from string
-    return parseSync(getSvg((symbols as any)[symbolName], {
+  if (symbolName) {
+    const svg = parseSync(getSvg((symbols as any)[symbolName], {
       width: size.width,
       height: size.height,
-    }))
+    }));
+
+    // Filter out non-path elements and modify path colors
+    const pathElements = svg.children
+    .filter((child: any) => child.name === 'path' && child.attributes.fill !== 'green')
+    .map((path: any) => {
+      const currentStrokeWidth = Number.parseFloat(path.attributes['stroke-width'] || '0.02');
+      const newStrokeWidth = (currentStrokeWidth * 1.5).toString();
+      
+      return {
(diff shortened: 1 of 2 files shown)


src/lib/soup-to-svg.ts
```diff
@@ -147,12 +147,57 @@ function createSchematicComponent(
...
+      
+      return {
+        ...path,
+        attributes: {
+          ...path.attributes,
+          stroke: path.attributes.stroke === 'black' ? 'red' : path.attributes.stroke,
+          'stroke-width': newStrokeWidth
+        }
+      };
+    });
```
seveibar, line 171: can you run biome on these files, we should add the format workflow but i'm noticing future reviews are getting messed up whenever someone isn't biome'ing with the latest config

I think this approach is too complicated because the code path is really long, in your approach you're doing:

```
symbol data -> svg -> parse svg into json -> modify svg json -> stringify svg
```

can we just implement:

```
symbol data -> svg
```

I know there's a bit of code overlap between the two repos, but we're just parsing `SchSymbol.primitives.filter(p => p.type === "path")

I think this will make transform issues a lot less confusing as well, you won't need to do logic to figure out if a path is green etc.

seveibar (changes requested):
i think this approach is going to be more fragile than just doing symbol data directly

src/utils/soup.json
```diff
@@ -60,10 +60,7 @@
     "source_port_id": "source_port_0",
```
seveibar, line 60: we should "ignore" this file (and maybe every json file) in the `biome.json`

seveibar:
FWIW this is great work, I just think we'd regret this approach because the SVG standard is just too extensive/weird to maintain

src/utils/soup.json
imrishabh18: I have added the `symbol_name` in this json. So, I don't think we can ignore it

seveibar: Gotcha np!

src/lib/soup-to-svg.ts
imrishabh18: Yeah, this makes sense. I will take a look at this approach

imrishabh18:
@anas-sarkez Are you sure the issue you mentioned is related to this PR? 🤔

seveibar (approved):

PR #30, opened 2024-08-30
cleaned console logs 

anas-sarkez:
console is filled with SVG object data


ShiboSoftwareDev (approved):
@imrishabh18 you left some console logs, did you want them to stay there?

Issue #49, opened 2024-09-06
Bounds calculation not correct with example file (some pads out of bounds)


seveibar:
![image](https://github.com/user-attachments/assets/0f0fac64-3d04-4c83-af60-45c7e537fd7f)

```json
[
  {
    "type": "source_port",
    "source_port_id": "source_port_0",
    "name": "pin1",
    "pin_number": 1,
    "port_hints": [
      "pin1",
      "1"
    ],
    "source_component_id": "source_component_0"
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_1",
    "name": "pin2",
    "pin_number": 2,
    "port_hints": [
      "pin2",
      "2"
    ],
    "source_component_id": "source_component_0"
  },
  {
    "type": "source_component",
    "source_component_id": "source_component_0",
    "ftype": "simple_chip",
    "name": "S1"
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_2",
    "name": "pin1",
    "pin_number": 1,
    "port_hints": [
      "pin1",
      "1"
    ],
    "source_component_id": "source_component_1"
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_3",
    "name": "pin2",
    "pin_number": 2,
    "port_hints": [
      "pin2",
      "2"
    ],
    "source_component_id": "source_component_1"
  },
  {
    "type": "source_component",
    "source_component_id": "source_component_1",
    "ftype": "simple_chip",
    "name": "S2"
  },
  {
    "type": "schematic_component",
    "schematic_component_id": "schematic_component_0",
    "center": {
      "x": 0,
      "y": 0
    },
    "rotation": 0,
    "size": {
      "width": 0.4,
      "height": 0.4
    },
    "pin_spacing": 0.2,
    "source_component_id": "source_component_0"
  },
  {
    "type": "schematic_component",
    "schematic_component_id": "schematic_component_1",
    "center": {
      "x": 0,
      "y": 0
    },
    "rotation": 0,
    "size": {
      "width": 0.4,
      "height": 0.4
    },
    "pin_spacing": 0.2,
    "source_component_id": "source_component_1"
  },
  {
    "type": "schematic_port",
    "schematic_port_id": "schematic_port_0",
    "schematic_component_id": "schematic_component_0",
    "center": {
      "x": 0,
      "y": 0
    },
    "source_port_id": "source_port_0",
    "facing_direction": "up"
  },
  {
    "type": "schematic_port",
[...]

PR #56, opened 2024-09-09
fix: Bounds calculation

abhijitxy:
#49


imrishabh18:
Also, have you made some changes in the bound calculation function, cause currently I can only see a test file added?

abhijitxy:
> Also, have you made some changes in the bound calculation function, cause currently I can only see a test file added?

 I haven't made direct changes to the bound calculation function yet, it is just a test file with some changes in the JSON

seveibar:
@abhijitxy i think in your PR title you shouldn't say "fix: " unless you're fixing the issue, it's totally ok to introduce a snapshot test in one PR then fix in the next PR but i would be explicit. If you open a draft PR it won't automatically tag people for review, so that's a good way to indicate if it's not ready to review

i'm a bit confused on the snapshot because it looks like the bounds calculation is correct? The original issue shows a failed bounds calculation. Maybe we can close this PR and the issue because it seems like there isn't a bug? Could be wrong?

image

PR #69, opened 2024-09-15
implemented colored fabrication note path/text

ShiboSoftwareDev:
don't we need to implement this in pcb-viewer?


imrishabh18 (approved):

imrishabh18:
I think the fabrication is already present in `pcb-viewer`

ShiboSoftwareDev:
yes the color

Issue #53, opened 2024-09-09
Draw `pcb_via`


seveibar:
## Should Look Like

![image](https://github.com/user-attachments/assets/a1538696-6483-41b5-84b9-015c07d92f1b)

## Currently Looks Like

![image](https://github.com/user-attachments/assets/56b26b3c-9c2b-4afc-b57c-5d7f0ad057a1)

```json
[
  {
    "type": "source_port",
    "source_port_id": "source_port_0",
    "name": "pin1",
    "pin_number": 1,
    "port_hints": [
      "-",
      "left",
      "pin1",
      "1"
    ],
    "source_component_id": "source_component_0"
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_1",
    "name": "pin2",
    "pin_number": 2,
    "port_hints": [
      "+",
      "right",
      "pin2",
      "2"
    ],
    "source_component_id": "source_component_0"
  },
  {
    "type": "source_component",
    "source_component_id": "source_component_0",
    "ftype": "simple_resistor",
    "name": "R1",
    "resistance": 100000
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_2",
    "name": "pin1",
    "pin_number": 1,
    "port_hints": [
      "anode",
      "pos",
      "pin1",
      "1"
    ],
    "source_component_id": "source_component_1"
  },
  {
    "type": "source_port",
    "source_port_id": "source_port_3",
    "name": "pin2",
    "pin_number": 2,
    "port_hints": [
      "cathode",
      "neg",
      "pin2",
      "2"
    ],
    "source_component_id": "source_component_1"
  },
  {
    "type": "source_component",
    "source_component_id": "source_component_1",
    "ftype": "simple_diode",
    "name": "LED1"
  },
  {
    "type": "source_trace",
    "source_trace_id": "source_trace_0",
    "connected_source_port_ids": [
      "source_port_1",
      "source_port_2"
    ],
    "connected_source_net_ids": []
  },
  {
    "type": "schematic_component",
    "schematic_component_id": "schematic_component_0",
    "center": {
      "x": 0,
      "y": 0
    },
    "rotation": 0,
    "size": {
      "width": 1,
      "height": 0.55
    },
    "source_component_id": "source_component_0",
    "symbol_name": "boxresistor_horz"
  },
  {
    "type": "schematic_component",
    "schematic_component_id": "schematic_component_1",
    "center": {
      "x": 0,
      "y": 0
    },
[...]

PR #72, opened 2024-09-16
feat: trace added with route_type via

imrishabh18:
Closes - #53 

Screenshot 2024-09-17 at 12 03 11 AM


seveibar (changes requested):
ack should have said this, but the correct behavior is to render `pcb_via` elements but _not_ render `route_type: "via"`- it's a bit confusing but if not there will be a lot of "double renders", there is currently an issue in core right now with not creating `pcb_via` elements even if the `route_type` has vias

PR #102, opened 2024-10-29
Update the documentation in readme.md

ni9999:
Right now the usage in readme is outdated and doesn't work anymore. This PR will update the usage codes in readme.


imrishabh18:
Thanks! @ni9999

PR #103, opened 2024-11-05
fix schematic component rendering relying on non-standard props

seveibar:


tests/sch/__snapshots__/kicad-theme-demo.snap.svg
```diff
@@ -9,4 +9,4 @@
               .pin-number { font-size: 8.490566037735848px; fill: rgb(169, 0, 0); }
```
seveibar, line 9: this regression is fixed when core is updated, but there's a chicken and egg issue

PR #106, opened 2024-11-06
feat: Add netalias

imrishabh18:
Closes - #91 

- Refactor stories directory structure


lib/sch/svg-object-fns/create-svg-objects-for-sch-net-label.ts
```diff
@@ -0,0 +1,123 @@
...
+        type: "text",
+        value: schNetLabel.text || "",
+        name: "",
+        attributes: {},
+        children: [],
+      },
+    ],
+    value: "",
+  })
+
```
seveibar, line 121: Can you make it more clear what the coordinate space is? Especially when youre in screen space “screenTextX” is much better. Also add the suffix Mm to indicate something is in milimeters (real space), eg labelWidthMm

seveibar (commented):
Looks good but i think could be improved with the renaming

PR #115, opened 2024-11-10
Use the source port name for the pin text

imrishabh18:
This will have the normalized pin labels


seveibar (changes requested):
Something is strange in the diff, looks like a regression, you probably need to do “bun i” can you show what this does in a snapshot? Pin numbers are supposed to appear outside the chip, same as kicad

imrishabh18:
Ohh yeah, forgot to do the `bun i`. There is no diff now

PR #116, opened 2024-11-10
Revert "Use the source port name for the pin text"

seveibar:
Reverts tscircuit/circuit-to-svg#115


seveibar:
@imrishabh18 this comment was not addressed 
image

imrishabh18:
@seveibar You meant the normalized pinLabels shouldn't be on top of the ports? It should always be pin number?

imrishabh18:
Gotcha, I didn't had the latest `core` got your point

PR #117, opened 2024-11-10
Nonexistent Symbol Handling, Bad Symbol Port Match Handling

seveibar:
- **add nonexistent symbol handling**
- **fix snapshot**
- **add error message when symbol ports don't align**


tests/sch/rotated-resistor.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+      circuit.getCircuitJson().map((c) => {
+        if (c.type !== "schematic_component") return c
+
+        return {
+          ...c,
+          symbol_name: "boxresistor_vert",
+        }
+      }),
+    ),
+  ).toMatchSvgSnapshot(import.meta.path)
```
seveibar, line 30: the ports fail to match for `box_resistorvert` because core doesn't properly computed rotated symbols

Issue #122, opened 2024-11-13
Fix left and right side of resistor port being drawn differently


seveibar:
Likely relevant file: https://github.com/tscircuit/circuit-to-svg/blob/main/lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-symbol.ts#L201-L202

![Image](https://github.com/user-attachments/assets/6400be35-9932-4751-8904-501c79207ee8)

AkashJana18:
How do i test it?
I couldn't find the contribution guide either to check how can i test it
Can you please guide @seveibar

imrishabh18:
@AkashJana18 To test the specific file use this command `bun test <path-of-test-file>`. This will create a snapshot of the file in the `__snapshots__` directory with the file name `.snap.svg`. And if there is some difference in the image it will create a `.diff.png`.

PR #131, opened 2024-11-16
[Fix] : Left and right side port of resistor

AkashJana18:
/claim #122 

Before : 
image

After:
image

image


lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-symbol.ts
```diff
@@ -199,10 +197,14 @@ export const createSvgObjectsFromSchematicComponentWithSymbol = ({
 
   // Draw Ports for debugging
   for (const port of symbol.ports) {
+    console.log("Port Object:", port); // Debug: Log each port
```
seveibar, line 200:
```suggestion
```

seveibar:
we really don't like magic numbers, so this solution probably isn't acceptable

seveibar (changes requested):
no magic numbers allowed. What is the underlying cause of the offset on only one side?

AkashJana18:
> we really don't like magic numbers, so this solution probably isn't acceptable

Okay, I will be right back with better solution

AkashJana18:
Hey @seveibar ,

After investigating the issue, I’ve identified that the root cause lies in the asymmetry of the port positions defined in the symbol itself. The slight misalignment in the coordinates causes the observed offset.
image


I haven’t applied any fixes yet, but addressing this at the source (by ensuring symmetrical and consistent port positions in the symbol definition) would be the cleanest solution, avoiding the need for workarounds or hacks in the code.

Looking forward to your feedback on this approach!

imrishabh18:
@AkashJana18 Great! I had this hunch that probably it will be on the symbol alignment but great that you showed the coordinates to prove it. It would be great if you can fix this in the schematic symbols.

AkashJana18:
Can you @imrishabh18 please tag the file ?

imrishabh18:
https://github.com/tscircuit/schematic-symbols/blob/main/assets/symbols/boxresistor.svg

This is the file which you can edit in inkscape tool, and align the two ports. You can then see the documentation on how to generate the files after you have edited and preview.

AkashJana18:
So i have to download the inkscape tool to edit the boxresistor.svg and push it right @imrishabh18

PR #132, opened 2024-11-17
Implemented create svg schematic text function 

anas-sarkez:


lib/sch/convert-circuit-json-to-schematic-svg.ts
```diff
@@ -120,8 +122,13 @@ export function convertCircuitJsonToSchematicSvg(
     } else if (elm.type === "schematic_trace") {
       schTraceSvgs.push(...createSchematicTrace(elm, transform))
     } else if (elm.type === "schematic_net_label") {
+      console.log(elm, transform)
```
seveibar, line 125:
```suggestion
```

lib/sch/convert-circuit-json-to-schematic-svg.ts
```diff
@@ -120,8 +122,13 @@ export function convertCircuitJsonToSchematicSvg(
...
       schNetLabel.push(...createSvgObjectsForSchNetLabel(elm, transform))
     }
+    else if (elm.type === "schematic_text") {
+      console.log(elm, transform)
```
seveibar, line 129:
```suggestion
```

lib/sch/svg-object-fns/create-svg-objects-for-sch-text.ts
```diff
@@ -0,0 +1,66 @@
...
+  },
+  transform: Matrix
+): SvgObject => {
+
+
+  // Apply transformation
+  const center = applyToPoint(transform, elm.position);
+
+  console.log("Transformed Center Position:", center);
+  console.log("Element Position:", elm.position);
```
seveibar, line 24: Check for logs

seveibar (changes requested):
This PR is _great_, but for prs introducing this type of function the snapshot test _must_ be rendered with grid with some labeled points (check out the net label snapshot) because if we dont check the points we often find out that the svg is off by a bit.

after that looks good!!!

anas-sarkez:
@seveibar checkout the snapshot

seveibar (approved):
Make sure to run format

PR #133, opened 2024-11-17
updated circuit-to-svg to stop drawing the manufacturer part numbers

anas-sarkez:


seveibar (approved):
can you update core, if you update core, we should still see the manufacturer numbers and there won't be a regression right?

seveibar (approved):
AWESOME

PR #144, opened 2024-12-07
feat: `schematic_net_label.symbol_name`

Abse2001:
created fn to create symbols using schematic_net_label.symbol_name
hard-coding "ground_horz" as symbol_name for now until I finish the core side


seveibar (commented):
Is there a way to avoid repeating so much code? Creating utility functions maybe?

seveibar:
It’s copied from the other symbol rendering code right? Can we put it into shared functions in a clean way?

seveibar (approved):
With the snapshot testing im not too worried about refactoring later i suppose

Abse2001:
@seveibar I tried to use the function that draw symbols but it take s schematic component not schematic label then I tried to convert the schematic net label to a schematic component but I ran to a problem it always say could not match ports,
if you now how to fix that I could use it

Abse2001:
Putting it into a shared function sounds good too

seveibar:
Lets do it in a future pr sinxe this is blocking @Abse2001

PR #153, opened 2024-12-21
Added  [sch_componet_name and MPN] to createSvgObjectsFromSchematicComponentWithBox fn

AnasSarkiz:


tests/sch/transform-attribute.test.tsx
```diff
@@ -18,9 +18,11 @@ test("svg should have data-real-to-screen-transform attribute", () => {
 
   // Extract the transform value and verify it's a valid matrix string
   const match = svg.match(/data-real-to-screen-transform="([^"]+)"/)
-
-  // @ts-ignore
-  expect(match[1]).toMatchInlineSnapshot(
-    `"matrix(54.5454545455,0,0,-54.5454545455,600,300)"`,
-  )
+  expect(match).not.toBeNull() // Ensure match exists before accessing [1]
+  if (match) {
```
seveibar, line 22: this is not the correct fix here, you're basically saying "sometimes don't test this"

seveibar: i would revert and do `bun test --update-snapshots test/path` so that i can see how it changed

seveibar (commented):
well the text is being created twice now right- so you'll want to conditionally avoid rendering it in the second place

tests/sch/transform-attribute.test.tsx
AnasSarkiz: I've fixed it with upgrades, bun

AnasSarkiz:
> well the text is being created twice now right- so you'll want to conditionally avoid rendering it in the second place

@seveibar  

``` tsx
else if (elm.type === "schematic_text") {
      const schComponents= su(circuitJson).schematic_component.get(elm.schematic_component_id)
      if(!schComponents) {
        schText.push(createSvgSchText(elm, transform))
      }
    }
```

seveibar:
Yea something like that @AnasSarkiz, although i think the existence of the id is probably enough- no need to search for the schComponent which is a bit expensive computationally

AnasSarkiz:
> Yea something like that @AnasSarkiz, although i think the existence of the id is probably enough- no need to search for the schComponent which is a bit expensive computationally

@seveibar 

``` tsx
 if (!elm.schematic_component_id.includes("schematic_component")) {
        schText.push(createSvgSchText(elm, transform))
      }
      
```

seveibar:
@AnasSarkiz that doesn't really make sense, just the existence of `elm.schematic_component_id` i.e. `if (elm.schematic_component_id)` means that it should be rendered within the schematic component group

AnasSarkiz:
@seveibar 

``` tsx
if (elm.type === "schematic_text" && !elm.schematic_component_id) {
      schText.push(createSvgSchText(elm, transform))
    }

seveibar:
Maybe? Id have to see it in the pr....

AnasSarkiz:
@seveibar 
already did

Screenshot 2024-12-23 at 7 35 20 AM

seveibar (approved):

Issue #154, opened 2024-12-15
Last recent updates text anchors are not acting as expected


Abse2001:
you can repro this in core update schematic-symbols and a lot of  tests will fail and give you these wrong text positions
![Image](https://github.com/user-attachments/assets/6391b188-c499-42df-af45-e93e2601ffe5)
![Image](https://github.com/user-attachments/assets/6f366b82-b04b-456e-91cf-057035b3a20d)

PR #155, opened 2024-12-22
fixed symbol text position to adjust for large schematic symbols

ShiboSoftwareDev:
for the missing symbols, I need to update core first, however it needs the schematic-symbols repo updated and also this repo for testing. I tested with yalc and it worked, I'll get back and fix the tests after core is merged.

/claim #154
image


lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-symbol.ts
```diff
@@ -173,21 +173,57 @@ export const createSvgObjectsFromSchematicComponentWithSymbol = ({
...
+      const transformScale = Math.abs(transform.a)
+
+      if (anchor.includes("bottom")) {
+        return baseOffset * transformScale
+      }
+      if (anchor.includes("top")) {
+        return -baseOffset * transformScale
+      }
+      return 0
+    }
```
seveibar, line 203: Does this function need to be nested?

seveibar (changes requested):
Can you use pkg.pr.new with your core branch?

ShiboSoftwareDev:
> Can you use pkg.pr.new with your core branch?

I don't know what that means

lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-symbol.ts
ShiboSoftwareDev: it doesn't I'll flatten it

seveibar:
Make a pr to core with your required core changes, then look at the actions. One of the actions should give you an installation command

tests/sch/transform-attribute.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { test, expect } from "bun:test"
 import { convertCircuitJsonToSchematicSvg } from "lib"
 
-test("svg should have data-real-to-screen-transform attribute", () => {
+test.skip("svg should have data-real-to-screen-transform attribute", () => {
```
seveibar, line 4: Dont skip this, just run bun upgrade

ShiboSoftwareDev: I forgot to remove the skip

ShiboSoftwareDev:
@seveibar I'm using the stackblitz version of core, what now?

seveibar:
@ShiboSoftwareDev shouldnt using your branch via stackblitz fix the snapshots? They still look broken

seveibar:
We're going to stop allowing broken snapshots is the tldr

seveibar:
Another way to avoid the dance is to make it so schematic symbols is backwards compatible, whoever fid the resistor thing made a change that wasnt backwards compatible

ShiboSoftwareDev:
they got fixed except one, let me double check

ShiboSoftwareDev:
@seveibar the rotated resistor test was strange, other than that all looks good

ShiboSoftwareDev:
@seveibar what happens now?

seveibar (approved):

PR #161, opened 2025-01-10
fix make transparent label point

### LINKED ISSUE #160: Net labels should not obscure schematic traces
Best fix is probably to draw the traces over the net labels, but we could also make the netlabel background somewhat transparent, we should see which approach looks OK

Will need to create a good test, I recommend porting this test: https://github.com/tscircuit/core/blob/main/tests/components/normal-components/chip-complex-crossings-simplified-1.test.tsx

Image

/bounty $10

techmannih:
fixes #160 
/claim #160


Anshgrover23 (changes requested):
Your bun test is failing, kindly update the snapshot.

Anshgrover23 (changes requested):
@techmannih also you have not added test according to issue description.

ShiboSoftwareDev (approved):
looks good but
@seveibar if you notice some debugging text is not there anymore, should we investigate this?
image

seveibar (changes requested):
can you make a test that shows what it looks like with traces going through it? you can render a circuit with a lot of traces by importing @tscircuit/core

seveibar:
@ShiboSoftwareDev yes that is strange and could cause breakage elsewhere- it wasn't this PR that did that was it? very strange- would bounty that issue if the bug is indeed real

seveibar:
@techmannih it looks ok... i kind of feel like we should go w/ 50% transparency or something, i might just have to play with it. If you show a couple screenshots of different transparencies that could help- it doesn't look super great right now

Anshgrover23 (changes requested):
@techmannih nice examples.

Abse2001:
@Anshgrover23 after it merge request a change "good work" wdyt?

seveibar (approved):
GOOD SHOW

PR #166, opened 2025-01-14
Support 3d Rendering via 3d-viewer with SvgRenderer

### LINKED ISSUE #138: Experiment: Support 3d Rendering via 3d-viewer with SvgRenderer
I think the first phase is to try this out and see the impact on bundle size

/bounty $40

DrSensor:
close #138
/claim #138

> [!WARNING]
> This draft has some sorcery. (review with caution!)
> if Copilot (or other AI) hallucinated, try: git blame, https://githistory.xyz, or https://semanticdiff.com

> [!IMPORTANT]
> server side usage require DOM emulation

usage
```sh
bun add circuit-to-svg @tscircuit/3d-viewer @happy-dom/global-registrator
```
```js
import { GlobalRegistrator } from "@happy-dom/global-registrator"
GlobalRegistrator.register()

import { CircuitJsonTo3dSvg } from "circuit-to-svg/3d"
```


DrSensor:
gonna create an issue on how to refactor @tscircuit/3d-viewer so the renderer can be swapped

seveibar:
@techmannih no need to review for things like conflicts, linting etc. while a PR is in draft! drsensor is a smart guy i think he's still putting together the PR

seveibar:
This is looking great, before we merge we may want to do a quick bundle size comparison. I'm also wondering if we can avoid the patches to the libraries.

DrSensor:
the patch in `three-stdlib` is trivial so I think we can create a PR to their repo. (somehow they use IIFE)

as for `@react-three/fiber`, I think there is something wrong with `bun`
https://bun.sh/docs/runtime/modules#importing-packages

hopefully they don't hardcode React stuff as special case 😅

---
oh wait, this might work. **bundle it before running the `bun test`**. the caveat is all `./tests/**.test.*` must be refactored (just need to change import statement to import from `./dist/index.js`)

seveibar:
Not 100% sure I understand the issue with react three fiber, but we could explicitly change the import to whatever import we need to avoid Bun's resolution right? I might be confused

image

seveibar:
Oh another random note, one way we can avoid messing up servers that use circuit-to-svg etc. is by having this being another export, so we could have people who want to use 3d stuff do `circuit-to-svg/3d`

That would be nice because we do render a lot of svgs on the server for snippets!

seveibar:
@DrSensor curious how this PR is going, we're using SVGs in all of our docs now and I think it'd be sick to have 3d as a selectable option- i would even default to 3d being the default tab a lot of the time

https://docs.tscircuit.com/

image

DrSensor:
it should be okay now to use on server-side despite some bundler trick in vite.config.js 😅
the next step would be working on https://github.com/tscircuit/3d-viewer/issues/115 to remove the github specifier in 👇🏽 along with hard-coded change in 3d-viewer
https://github.com/tscircuit/circuit-to-svg/blob/fd2e2d7ca957a15c926054ea83a5a40dccc8a50c/package.json#L55-L57

Issue #165, opened 2025-01-12
Allow showing PCB Rats Nest for PCB View


seveibar:
Should be a flag when converting

The pcb viewer already supports rats nest, can be used as a reference

![Image](https://github.com/user-attachments/assets/20ed8028-91c7-467d-93a4-2f6dfaf5859d)

CC @ShiboSoftwareDev 

/bounty $10

PR #167, opened 2025-02-11
feat: rats nest

ArnavK-09:
/claim #165


lib/pcb/svg-object-fns/create-svg-objects-from-pcb-rat-nests.ts
```diff
@@ -0,0 +1,149 @@
...
+  transform: Matrix,
+): SvgObject[] {
+  // Build a lookup map by cleaned ID for any element that might provide a position.
+  const elementMap = new Map<string, AnyCircuitElement>()
+  for (const elm of soup) {
+    const id =
+      (elm as any).id || (elm as any).pcb_port_id || (elm as any).source_port_id
+    if (id) {
+      const cleanedId = id.replace(/_\d+$/, "")
+      elementMap.set(cleanedId, elm)
```
seveibar, line 24: the ids aren't always going to be in the format `source_port_...`, this code will fail in weird ways (in general, it's best to avoid regex'ing or string manipulation)

I think you can look at the connectivity map to find all the connected ids then iterate over them instead?

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-rat-nests.ts
```diff
@@ -0,0 +1,149 @@
...
+        const dy = sourcePoint.y - pos.y
+        const distance = Math.sqrt(dx * dx + dy * dy)
+        if (distance > 0 && distance < minDistance) {
+          minDistance = distance
+          nearestPoint = pos
+        }
+      }
+    }
+    return nearestPoint
+  }
```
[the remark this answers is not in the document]
seveibar: the reason it's "rats nest" it it's a rat's nest. The apostrophe is omitted hence the weird grammar 😁

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-rat-nests.ts
```diff
@@ -0,0 +1,149 @@
...
+    return []
+  }
+
+  // Filter for ports and traces that are relevant for rats nest.
+  const pcbPorts = soup.filter((elm) => elm.type === "pcb_port")
+  const sourceTraces = soup.filter((elm) => elm.type === "source_trace")
+  const sourcePorts = soup.filter((elm) => elm.type === "source_port")
+
+  const getSourcePortForPcbPort = (pcbPortId: string) => {
+    return sourcePorts.find((elm) => (elm as any).pcb_port_id === pcbPortId)
```
seveibar, line 55: Can you use `@tscircuit/soup-util` for this? Yes the name is weird, but it has much much better types and also checks for element types properly and auto-optimizes

```suggestion
    return su(circuitJson).source_port.getWhere({ pcb_port_id })
```

Generally you want to use soup-util instead of scanning the circuit json because it'll auto-optimize and not get slow even as the array gets massive. It has a similar interface to a db client

https://github.com/tscircuit/soup-util

seveibar (commented):
nice work, looking good. I think the id method you're using will be pretty buggy, so that's the main thing. Those id strings can change quite a bit in production

You could also load in a more complicated rats nest (literally the same one that the PCB viewer has here: https://github.com/tscircuit/pcb-viewer/blob/main/src/stories/trace-and-net.stories.tsx) and that would show this was implemented totally properly

+$15 if you're able to create a test with that exact circuit

NOTE: You can add @tscircuit/core as a dev dependency and literally create the circuit json in the exact same way that @ShiboSoftwareDev did- these are usually a bit more readable than just having circuit json, although both types of tests (ones that use circuit json, and ones that use core) are very good to have!!

tests/pcb/pcb-rats-nests.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <trace path={[".R3 > .left", ".R2 > .right"]} />
+  </board>,
+)
+
+const soup = circuit.getCircuitJson()
+
+test("shouldDrawRatsNest false", () => {
+  expect(convertCircuitJsonToPcbSvg(soup)).toMatchSvgSnapshot(
+    import.meta.path,
+    "pcb-trace-error-shouldDrawRatsNest-false",
```
seveibar, line 49: hehe you don't need to test the false case again here

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-rats-nests.ts
```diff
@@ -0,0 +1,109 @@
...
+      line.endPoint.x,
+      line.endPoint.y,
+    ])
+    const attributes: { [key: string]: string } = {
+      x1: transformedStart[0].toString(),
+      y1: transformedStart[1].toString(),
+      x2: transformedEnd[0].toString(),
+      y2: transformedEnd[1].toString(),
+      stroke: "white",
+      "stroke-width": "1",
```
seveibar, line 94: so the nets are supposed to be dotted

seveibar: do you see how shibo has both dotted and solid, i think it would be trivial to make the ones connected by net dotted yea?

lib/pcb/create-svg-objects-from-pcb-rats-nest/get-source-port-for-pcb.ts
```diff
@@ -0,0 +1,9 @@
+import type { AnyCircuitElement } from "circuit-json"
+import { su } from "@tscircuit/soup-util"
+
+export const getSourcePortForPcbPort = (
+  pcbPortId: string,
+  soup: AnyCircuitElement[],
+) => {
+  return su(soup).source_port.getWhere({ pcb_port_id: pcbPortId })
```
seveibar, line 8: so when you have a function that is so simple that it only has a line or two, you actually increase the complexity of the codebase by increasing the "go-to-definition-depth" (or "complexity depth") of the code, so you should remove this function and call `su` directly from the call site

lib/pcb/create-svg-objects-from-pcb-rats-nest/get-element-position.ts
```diff
@@ -0,0 +1,19 @@
...
+      (element as any).pcb_port_id ||
+      (element as any).source_port_id
+    return elmId === id
+  })
+
+  if (elm && "x" in elm && "y" in elm) {
+    return { x: (elm as any).x, y: (elm as any).y }
+  }
+  return null
+}
```
seveibar, line 19: do you still need this? It's very hacky and won't work correctly because it doesn't filter based on elm.type

seveibar (approved):
/tip $15 amazing work!!

PR #181, opened 2025-02-27
Remove classnames for silkscreen

seveibar:
@techmannih can you remove all of these classnames? They are causing the stroke bug i spoke about


seveibar:
there should be no classnames defined like this at all!

PR #182, opened 2025-02-27
remove classname for silkscreen

techmannih:
remove class name for silkscreen


seveibar (changes requested):
don't break anything, so adapt the code to work without the classnames

seveibar:
this is super broken, why request review?

Maybe try removing one class name at a time. You might be overrelying on AI, there are completely unnecessary rewrites- it's just a style change

PR #184, opened 2025-02-27
fix silkscreen text style

seveibar:


seveibar:
CC @techmannih

seveibar:
> @seveibar oh like that, now there is need to remove other classname?

They all must go! @techmannih

PR #193, opened 2025-03-11
add support pill shape hole with rect plated

techmannih:
add support pill shape hole with rect plated


seveibar (approved):
going to let @imrishabh18 leave a review as well

imrishabh18 (approved):

PR #199, opened 2025-04-21
fixed automatic dark mode

### LINKED ISSUE #197: Improve automatic dark mode colors
Image

/bounty $8

There are lots of ways to turn on automatic dark mode. I use the "Dark Reader" chrome extension, but you can also enable via chrome://flags

ShiboSoftwareDev:
default (when overriding some colors):
image

automatic dark mode:
image

/claim #197
/closes #197


tests/sch/__snapshots__/kicad-theme-demo.snap.svg
```diff
@@ -1,8 +1,8 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="1200" height="600" style="background-color: rgb(245, 241, 237)" data-real-to-screen-transform="matrix(65.9340659341,0,0,-65.9340659341,423.0920010989,303.2967032967)"><style>
```
seveibar, line 1: regression? The component body should not be transparent.

Also it's pretty hard to read these snapshots on Github now, and we use these snapshots everywhere. Can you figure out a way to make it work without making snapshots hard to read? You can always make the background a parameter and we can set it to "transparent" on tscircuit.com etc. The default should be readable on github.

seveibar:
seems good, but confused about component body and I think we should default to readable on github (we can use an option in the schematic viewer repo to configure the bg to be transparent, perhaps we could make it so that the user can reconfigure _any_ of the colors in the color object, this way we could also change the component body.

ShiboSoftwareDev:
@seveibar I can instead pass down mergedColorMap directly, I'm not sure which is the better approach

lib/sch/svg-object-fns/create-svg-objects-for-sch-net-label.ts
```diff
@@ -23,44 +24,55 @@ import {
   END_PADDING_EXTRA_PER_CHARACTER_FSR,
 } from "../../utils/net-label-utils"
 
-export const createSvgObjectsForSchNetLabel = (
-  schNetLabel: SchematicNetLabel,
-  realToScreenTransform: Matrix,
-): SvgObject[] => {
-  if (!schNetLabel.text) return []
+export const createSvgObjectsForSchNetLabel = ({
+  elm,
```
seveibar, line 28: rename is worse

```suggestion
  elm: schNetElm,
```

lib/sch/svg-object-fns/create-svg-objects-for-sch-net-label.ts
```diff
@@ -124,7 +132,7 @@ export const createSvgObjectsForSchNetLabel = (
   ].map((fontRelativePoint) =>
     applyToPoint(
       compose(
-        realToScreenTransform,
+        transform,
```
seveibar, line 135: Rename is worse

seveibar (commented):
Yea you should computed the merged color map and pass it down as colorMap, repeating that operation is slow and prone to error

PR #200, opened 2025-04-25
relative border and text sizes

ShiboSoftwareDev:
relative to the size of the element

now even the small components appear clearly
image

/claim tscircuit/assembly-viewer#5
/closes tscircuit/assembly-viewer#5


seveibar (commented):
i don't think this looks better tbh, there are definitely scaling issues but every component have the same stroke width looks really weird i think

ShiboSoftwareDev:
@seveibar the was that the small components had large borders and that was fixed, how do you think I can improve the looks?

seveibar:
i'm not sure but we should work from a reference, so let's find assembly drawings that are very good

seveibar (approved):

ShiboSoftwareDev:
@seveibar unable to merge

PR #203, opened 2025-05-01
override any schematic color

ShiboSoftwareDev:


lib/sch/convert-circuit-json-to-schematic-svg.ts
```diff
@@ -32,8 +32,7 @@ interface Options {
 
 export interface ColorOverrides {
   schematic?: {
-    background?: string
-    component_body?: string
+    [key: string]: string
```
seveibar, line 35: aren't all the key values known? You can use `keyof typeof` to extract them from the color map

seveibar (changes requested):
should be easy to add amazing types here!

lib/sch/convert-circuit-json-to-schematic-svg.ts
ShiboSoftwareDev: true, how didn't I think of that XD

seveibar (approved):
Nice

PR #208, opened 2025-05-04
draw drc errors that involve vias

ShiboSoftwareDev:


seveibar (approved):
This is great, can we draw the bounding box around the error (is that possible?, or is there not enough info, maybe we can use the id’d elements?) can be done in next pr

ShiboSoftwareDev:
> This is great, can we draw the bounding box around the error (is that possible?, or is there not enough info, maybe we can use the id’d elements?) can be done in next pr

I would've merged and started another pr, but oh well I'll just patch

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-trace-error.ts
```diff
@@ -118,3 +178,92 @@ export function createSvgObjectsFromPcbTraceError(
...
+      {
+        name: "rect",
+        type: "element",
+        attributes: {
+          x: (midX - 5).toString(),
+          y: (midY - 5).toString(),
+          width: "10",
+          height: "10",
+          fill: "red",
+          transform: `rotate(45 ${midX} ${midY})`,
```
seveibar, line 237: is this angle supposed to be hard coded?

seveibar (approved):
excellent /tip $10

PR #210, opened 2025-05-04
add support solder paste for smtpads

### LINKED ISSUE #204: Support outputting top or bottom PCB stencils when configured with an option
Requires https://github.com/tscircuit/core/issues/771

/bounty $25

techmannih:
fixes #204 
/claim #204


seveibar (commented):
This is really great, but when we render solder paste the layer we render should be a specific option!

tests/pcb/pcb-solder-paste.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      type: "pcb_solder_paste",
+      x: 0,
+      y: 0,
+      layer: "bottom",
+      shape: "circle",
+      radius: (2 / 2) * 0.7,
+      pcb_solder_paste_id: "solder_paste_4",
+      pcb_component_id: "pcb_component_0",
+    },
+  ])
```
seveibar, line 66: Where is the option? Only render solder paste if a specific option is provided, it is not normally rendered

seveibar: I believe the issue describes the API but probably { mode: “solder_paste”, layer: “top” }

techmannih: Yeah update it

lib/pcb/layer-name-to-color.ts
```diff
@@ -12,3 +12,15 @@ export function layerNameToColor(layerName: string) {
...
+export const SOLDER_PASTE_LAYER_NAME_TO_COLOR = {
+  bottom: "rgb(169, 169, 169)", // Lighter gray (Light gray)
+  top: "rgb(105, 105, 105)", // Darker gray (Dim gray)
+}
+
+export function solderPasteLayerNameToColor(layerName: string) {
+  return (
+    SOLDER_PASTE_LAYER_NAME_TO_COLOR[
+      layerName as keyof typeof SOLDER_PASTE_LAYER_NAME_TO_COLOR
+    ] ?? "gray" // Gray color for solder paste
```
seveibar, line 24: I think we want to make it always black, on any layer, since you export the layers separately and theyre fed into a laser cutter machine

lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -47,6 +49,10 @@ interface Options {
   height?: number
   shouldDrawErrors?: boolean
   shouldDrawRatsNest?: boolean
+  mode?: {
```
seveibar, line 52: You have to also add a type for the regular mode, even if its not used.

i think the structure of this object is a bit confusing. What if we had a separate function convertCircuitJsonToSolderPasteMask(circuitJson, { layer: “top” })?

seveibar (changes requested):
Can you implement the approach with the separate function, tyty

lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -345,3 +356,30 @@ function createSvgObjectFromPcbBoundary(
...
+  const boardElements = soup.filter((elm) => elm.type === "pcb_board")
+
+  // Combine filtered elements
+  const filteredSoup = [...boardElements, ...solderPasteElements]
+
+  return convertCircuitJsonToPcbSvg(filteredSoup, {
+    width: options.width,
+    height: options.height,
+  })
+}
```
seveibar, line 385: this should not be in the same file

lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -248,6 +257,8 @@ function createSvgObjects(
       return createSvgObjectsFromPcbHole(elm, transform)
     case "pcb_smtpad":
       return createSvgObjectsFromSmtPad(elm, transform)
+    case "pcb_solder_paste":
+      return createSvgObjectsFromSolderPaste(elm, transform)
```
seveibar, line 261: we never want to draw the solder paste for pcb svgs

also the way you implemented `Options` here is not the tscircuit 2 parameter convention https://github.com/tscircuit/handbook/blob/main/guides/code.md#1-two-parameter-rule

seveibar (changes requested):
it's bad practice to have a file export multiple functions, especially when the file is already long

seveibar:
@techmannih this is super unconventional, you added a function as a parameter. You should have created a new function, in a new file

PR #211, opened 2025-05-05
Only display pin1 indicator for chips, now as a filled up triangle instead of a notch

ShiboSoftwareDev:


lib/assembly/convert-circuit-json-to-assembly-svg.ts
```diff
@@ -166,6 +166,7 @@ function createSvgObjects(
             transform,
             { x: firstPort.x, y: firstPort.y },
             sourceComponent.name,
+            sourceComponent.ftype,
```
seveibar, line 169: FWIW instead of stacking more position parameters onto this function, it should be converted to a two parameter function https://github.com/tscircuit/handbook/blob/main/guides/code.md#1-two-parameter-rule

lib/assembly/svg-object-fns/create-svg-objects-from-assembly-component.ts
```diff
@@ -16,55 +16,59 @@ export function createSvgObjectsFromAssemblyComponent(
...
   name?: string,
+  ftype?: string,
 ): SvgObject {
   const { center, width, height, rotation = 0, layer = "top" } = component
   const [x, y] = applyToPoint(transform, [center.x, center.y])
   const [pinX, pinY] = applyToPoint(transform, [firstPin.x, firstPin.y])
   const scaledWidth = width * Math.abs(transform.a)
   const scaledHeight = height * Math.abs(transform.d)
 
+  const isChip = ftype?.includes("chip")
```
seveibar, line 27: this isn't the right check, although this is definitely the right idea.

There is a new circuit json property called `are_pins_interchangeable`, that is what you should use. If something has interchangeable pins, it should not have a triangle.

See https://github.com/tscircuit/circuit-json/pull/180

tscircuit core does not currently generate this property

seveibar (commented):
this is good but can we use `are_pins_interchangeable`? That's also a good issue to pick up because core doesn't generate those pins currently

seveibar (approved):

lib/assembly/convert-circuit-json-to-assembly-svg.ts
```diff
@@ -160,14 +160,17 @@ function createSvgObjects(
 
       // Proceed only if both sourceComponent and firstPort are found
       if (sourceComponent && firstPort) {
-        return [
-          createSvgObjectsFromAssemblyComponent(
+        const isChip = sourceComponent.are_pins_interchangeable === undefined
```
seveibar, line 163: Bad bool name

seveibar: The original name arePinsInterchangeable would be good

seveibar:
@ShiboSoftwareDev minor but i would stay away from using "transform" as a var name because the transform direction is often unclear, it's better to say `realToScreenTransform` etc.

When you're transforming e.g. a mouse coordinate you use the inverse transform, `screenToRealTransform`, note that "Real" means real unit, i.e. `mm`

PR #220, opened 2025-05-20
updated dep

Abse2001:
there is definitely a regression happening because of the schematic trace calculation, I had to Increase the test time for two more tests.

CC @seveibar @ShiboSoftwareDev

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -22,7 +22,7 @@
         "@storybook/test": "^8.2.5",
         "@tscircuit/checks": "^0.0.44",
         "@tscircuit/circuit-json-util": "^0.0.47",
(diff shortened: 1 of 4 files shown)


ShiboSoftwareDev:
@Abse2001 how did you conclude it was because of the sch trace calculation? The previous timeout was caused by the pcb autorouter

seveibar (approved):
Glad you noted the perf regressiom. We should create a benchmark test to recreate the slow timing

Abse2001:
@ShiboSoftwareDev I'm assuming at this point but after updating core it happened in testes that has alot of traces you should do a deep investigation into the issue.

PR #221, opened 2025-05-22
support nine point anchor alignment for silkscreen texts

ShiboSoftwareDev:


imrishabh18 (approved):

seveibar (commented):
that snapshot test is not possible to understand at a glance because there are two different points (maybe 3?). I would break it into 3 tests.

seveibar (approved):
REALLY NICE!!!!!

PR #284, opened 2025-07-14
schematic table text same color as borders

ShiboSoftwareDev:


lib/sch/svg-object-fns/create-svg-objects-from-sch-table.ts
```diff
@@ -242,7 +242,7 @@ export const createSvgObjectsFromSchematicTable = ({
           "font-size": `${fontSize}px`,
           "text-anchor": textAnchorMap[horizontal_align],
           "dominant-baseline": dominantBaselineMap[vertical_align],
-          fill: colorMap.schematic.sheet_label,
+          fill: "#666",
```
seveibar, line 245: hmm shouldn't this be in the colorMap

Issue #294, opened 2025-08-08
Draw reference designators on schematic on top of schematic traces


seveibar:
<img width="1212" height="728" alt="Image" src="https://github.com/user-attachments/assets/793014bc-e4d8-4ca0-bdf4-366dc55c604f" />

seveibar:
<img width="2192" height="792" alt="Image" src="https://github.com/user-attachments/assets/dd451739-3229-4e33-adc2-bf2d4ef21435" />

seveibar:
the green line is appearing over the text (the reference designator)

PR #297, opened 2025-08-27
Draw reference designators on schematic on top of schematic traces

t431w:
Ref #294 

/claim #294

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "circuit-to-svg",
   "type": "module",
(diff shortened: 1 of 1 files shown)


seveibar (changes requested):
needs reproduction where a trace line goes through text

seveibar (changes requested):
see discord comment about new long name

seveibar (changes requested):
Im not sure about the readability, can you add an outline to the text via stroke

PR #300, opened 2025-08-30
feat: draw input/output arrows on schematic_port

ShiboSoftwareDev:
these will be present for example on `pinAttributes: { requiresPower }`


seveibar (approved):
very, very nice test

PR #303, opened 2025-09-04
feat: highlight all traces under the same net

imrishabh18:


seveibar:
@imrishabh18 yes even at that deployment link, it doesn't highlight the subcircuit when you hover over traces. It only appears to work in some examples

seveibar:
e.g. 
image


It also doesn't work when you download the 555 timer svg

PR #307, opened 2025-09-05
feat(pcb): Add support for drawing copper pours

ShiboSoftwareDev:
This pull request introduces support for rendering pcb_copper_pour elements in the PCB SVG output. It enables
the visualization of rectangular and polygonal copper pours on both top and bottom layers.

Changes

• New SVG Object Function: Created createSvgObjectsFromPcbCopperPour to handle the conversion of
pcb_copper_pour JSON elements into SVG objects.
• Supports rect and polygon shapes.
• rect pours can be rotated.
• PCB SVG Conversion: Updated convertCircuitJsonToPcbSvg to:
• Incorporate copper pours into the bounds calculation.
• Add pcb_copper_pour to the element processing switch case and rendering order.
• Testing: Added a new test file with snapshot tests to verify the correct rendering of rectangular, rotated
rectangular, and polygonal copper pours.


seveibar:
@ShiboSoftwareDev for some of the tests that test trace drawing, use `<board schMaxTraceDistance={5}` to make the traces be drawn

PR #308, opened 2025-09-08
feat(pcb): add support for brep copper pours

ShiboSoftwareDev:
- A new utility function `ringToPathD` has been created to convert B-rep rings into SVG path data.                                           
- The PCB SVG conversion logic now processes `pcb_copper_pour` elements with `shape: "brep"` and renders them as an SVG `<path>`.            
- The `fill-rule="evenodd"` attribute is used to ensure that inner rings are correctly rendered as holes in the pour.                        
- A comprehensive test case has been added to verify the rendering of various complex B-rep shapes, including polygons, rotated rectangles,  
and pours with multiple holes.


ShiboSoftwareDev:
I don't know why a potentiometer went missing in one test

seveibar (approved):
Nice!

PR #315, opened 2025-09-15
Add support for rendering soldermask part of SVG (if renderSolderMask : true)

nailoo:


tests/pcb/render-soldermask-option.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+test("renders soldermask overlay when enabled", () => {
+  const svg = convertCircuitJsonToPcbSvg(circuit, {
+    colorOverrides: {
+      copper: { top: "#ff0000" },
+      soldermask: { top: "#00ff00" },
+    },
+    renderSolderMask: true,
+  })
+  expect(svg).toMatchSvgSnapshot(import.meta.path, "enabled")
```
seveibar, line 40: Keep one test per file. In this case you can combine the snapshots together and add eg silkscreen to indicate whochbis which

The snapshot names arent good, this will be solved by combining them

PR #316, opened 2025-09-15
Title: fix(pcb): Correct arc rendering in BREP shapes

ShiboSoftwareDev:
This PR addresses an issue where arcs in BREP shapes, such as those in copper pours, were being  
drawn in the opposite direction. The incorrect rendering was due to the sweep-flag for the SVG arc path
command being inverted.                                                                                

The fix involves correcting the logic for the sweep-flag in lib/utils/ring-to-path-d.ts to account for 
the Y-axis flip in the SVG coordinate system. This ensures that arcs in BREP shapes are now drawn      
correctly.


ShiboSoftwareDev:
I noticed this when implementing in other repos

PR #322, opened 2025-09-16
 Render solder mask for covered SMT pads 

nailoo:

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-smt-pads.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-smt-pads.ts
@@ -1,12 +1,13 @@
 import type { PcbSmtPad } from "circuit-json"
 import { applyToPoint } from "transformation-matrix"
+import type { SvgObject } from "lib/svg-object"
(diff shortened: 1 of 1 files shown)


seveibar:
basically to make this robust you'd want to have each type of smtpad, including polygon smtpads, demonstrated to be working

lib/pcb/svg-object-fns/create-svg-objects-from-smt-pads.ts
```diff
@@ -11,122 +11,117 @@ export function createSvgObjectsFromSmtPad(
...
-        },
-      ]
+  const solderMaskColor =
+    colorMap.soldermask[pad.layer as keyof typeof colorMap.soldermask] ??
+    colorMap.soldermask.top
+
+  const createPadElements = (
+    elementName: "rect" | "circle" | "polygon",
+    geometry: Record<string, string>,
+  ) => {
```
seveibar, line 23: why do we need a closure?

seveibar (approved):
awesome /tip $20

PR #329, opened 2025-09-24
Fix the chip offsets

MustafaMulla29:
/claim [#789](https://github.com/tscircuit/tscircuit/issues/789)


imrishabh18 (approved):
Nice!

seveibar (changes requested):
introduces a new bug

Image

you have to conditionally render a longer line based on `port.is_disconnected` or whatever (the condition that does the circle or not)

seveibar (approved):

PR #345, opened 2025-10-03
Pinout: dynamic scaling for small pads, improved label separation

ShiboSoftwareDev:
• Make pinout diagrams more readable on small-pad boards by dynamically scaling UI elements and tightening SVG bounds.    
 • Improve label stacking/spacing so primary vs. other labels are visually separated.


seveibar (approved):

seveibar (changes requested):
Hold on, i just noticed there’s this uiScale parameter

seveibar:
Why is there uiScale? The pin label size is in mm, and everything should be in mm without scaling units

PR #351, opened 2025-10-04
remove component color and text for pinout

t431w:
ref https://github.com/tscircuit/tscircuit/issues/895


seveibar (changes requested):
CC @ShiboSoftwareDev (we should make you a codeowner)

I think this is not the right solution because it prevents any "chips on the board" from showing

The problem is maybe moreso with the test, @ShiboSoftwareDev imported a full pico which isn't correct (I didn't realize this), it needs to be the PICO design otherwise chips won't show

ShiboSoftwareDev:
@seveibar since that pico is a single chip isn't it expected to be a single component? This means there is no issue here if I understand correctly

PR #360, opened 2025-10-05
fix: change the color for  soldermask

nailoo:


lib/pcb/colors.ts
```diff
@@ -51,7 +51,7 @@ export const DEFAULT_PCB_COLOR_MAP: PcbColorMap = {
     bottom: "rgb(77, 127, 196)",
   },
   soldermask: {
-    top: "rgb(200, 52, 52)",
+    top: "rgb(18, 82, 50)",
```
imrishabh18, line 54: can you share from where did you get this color for solder mask? we need to follow the kicad color scheme.

PR #387, opened 2025-10-16
Add corner_radius support for rect SMT pads

techmannih:
ref https://github.com/tscircuit/tscircuit/issues/1077


seveibar (approved):
Yes but please also try to add silkscreen to the snapshot that describes what is happening

seveibar (approved):

PR #406, opened 2025-10-26
chore: remove useless gray box and text

jayantpranjal0:
ref https://github.com/tscircuit/tscircuit/issues/895
/claim https://github.com/tscircuit/tscircuit/issues/895


ShiboSoftwareDev:
The gray box turned out to not be a bug, since the pico is used as a single component, also you can't claim a multi-step PR in a single fix

ShiboSoftwareDev:
Even the text is correct

PR #409, opened 2025-10-29
refactor(pcb): Improve panel rendering and refactor numeric parsing 

ShiboSoftwareDev:
This pull request introduces support for rendering multi-board panels, refactors numeric       
parsing to use distance.parse, and fixes a related bug in bounds calculation.                  

Key Changes:                                                                                   

 1 Multi-Board Panel Support:                                                                  
    • The SVG bounding box calculation in both PCB and solder paste mask renderers now         
      correctly includes pcb_panel elements, allowing panels with multiple boards to be        
      rendered properly.                                                                       
    • pcb_panel is no longer considered a board for board-specific boundary calculations, which
      improves logical consistency.                                                            
    • A new test case and snapshot have been added to verify the visual output of a multi-board
      panel.                                                                                   
 2 Refactor to distance.parse:                                                                 
    • The custom toNumeric helper function has been completely removed.                        
    • All numeric parsing of distance values now uses distance.parse from the circuit-json     
      library. This standardizes how distance units are handled and reduces maintenance.       
 3 Bug Fix:                                                                                    
    • Fixed a crash that occurred when calculating the bounds for pcb_smtpad elements with a   
      polygon shape. The bounds calculation logic was reordered to handle polygon pads         
      correctly, preventing an error from distance.parse receiving NaN values.


lib/pcb/convert-circuit-json-to-solder-paste-mask.ts
```diff
@@ -12,6 +12,7 @@ import { createSvgObjectsFromSolderPaste } from "./svg-object-fns/convert-circui
 import type { PcbContext } from "./convert-circuit-json-to-pcb-svg"
 import { DEFAULT_PCB_COLOR_MAP } from "./colors"
 import { getSoftwareUsedString } from "../utils/get-software-used-string"
+import { toNumeric } from "../utils/to-numeric"
```
seveibar: this drives me nuts `toNumeric` is ALWAYS A BUG!

we do not pass around numbers as strings ever! It's not in any spec! use `distance.parse` for parsing strings that may be represented w/ mm etc.! Anything else is going to break!

PR #413, opened 2025-11-01
Dynamically rotate net label symbols based on anchor_side

### LINKED ISSUE #157: refactor create-svg-objects-for-sch-net-label-with-symbol.ts to render any symbol with the correct text anchors
the current code only works with ground symbol.

Exceluyi:
/claim #157
/closes #157

Description:
This PR refactors the `createSvgObjectsForSchNetLabelWithSymbol` function to correctly render any symbol with the appropriate rotation based on the `anchor_side` property of the `SchematicNetLabel`.

Previously, the symbol rotation was hardcoded to 0, causing incorrect orientation for symbols other than the default "ground" symbol. This change introduces dynamic rotation by utilizing the `getPathRotation` utility function, ensuring that symbols are displayed with the correct orientation relative to their net label's anchor side.

Additionally, misleading comments regarding symbol orientation and text placement within the symbol rendering logic have been removed for clarity and accuracy.

A new test case (`net-label-any-symbol-orientation.test.tsx`) was added to verify the correct rotation of a generic "ground" symbol for all four `anchor_side` values (left, right, top, bottom).


imrishabh18 (changes requested):
The issue mentions that `GND` text works, try with some other text anchor. First to reproduce the issue if has been fixed or not and then go for the fix

PR #423, opened 2025-11-14
feat: Adjust schematic voltage probe label alignment and arrow position

ShiboSoftwareDev:
This change enhances the rendering of schematic_voltage_probe labels for better alignment and clarity.             

 • Labels are now left-aligned for improved readability.                                                          
 • When a probe has both a name and a voltage, the arrow now points to the name, with the voltage appearing on the
   line below it.


seveibar (approved):
1. should be left-aligned
2. arrow should always be "coming from the name" if the name is defined, with the voltage below it

PR #424, opened 2025-11-16
Match voltage probe colors to simulation graph colors

ShiboSoftwareDev:
Summary
- Voltage probes now automatically use the same color as their corresponding simulation graphs when the probe name matches the graph name
- This provides better visual correlation between probes in the schematic and their waveforms in the simulation output

 Changes
- Added `buildGraphNameToColorMap()` function to build a mapping from simulation graph names to their assigned colors
- Updated schematic voltage probe renderer to accept an optional `graphColor` parameter
- Modified schematic-simulation combined SVG converter to pass graph colors to voltage probes
- Added comprehensive test to verify color matching behavior

 Test Coverage
- New test `voltage-probe-color-match.test.tsx` verifies that probes with names "vout", "vin", and "gnd" use the same colors as their matching simulation graphs


seveibar (changes requested):
why not determine the color in core? It will allow the user to customize it and keep it more consistent across viewers

PR #425, opened 2025-11-17
feat(sch, sim): Use color property for voltage probes and graphs

ShiboSoftwareDev:
Updates circuit-json and adapts the SVG renderers to use the new color property on voltage probes and simulation  
graphs.                                                                                                           

 • Schematic voltage probes now use their color property if provided.                                             
 • Simulation graphs will use their own color, fall back to the associated simulation_voltage_probe's color, and  
   then to the default palette.                                                                                   
 • Tests have been updated to reflect schema changes in circuit-json (e.g., using source_component_id).


seveibar (approved):
snapshot updates would be good, requires circuit json updates

PR #427, opened 2025-11-18
Add Support for Path-Based PCB Cutouts

Abse2001:
Implements SVG generation for pcb_cutout objects using the new path shape. Supports continuous and dashed slot patterns, corner radii, dimension validation, and full route-based geometry. Integrates with the existing cutout pipeline and adds snapshot tests for coverage.


seveibar (commented):
Wayyy too confusing

PR #435, opened 2025-11-24
fix: rendering of soldermask layer rendering for the whole board and the soldermask_margin

imrishabh18:
Added the substrate color around the smtpads to represent the margin gap around the copper. Not sure if it's right maybe  it should be black too?

image


lib/pcb/colors.ts
```diff
@@ -37,6 +42,8 @@ export interface PcbColorOverrides {
   silkscreen?: Partial<PcbColorMap["silkscreen"]>
   boardOutline?: string
   soldermask?: Partial<PcbColorMap["soldermask"]>
+  soldermaskCovered?: Partial<PcbColorMap["soldermaskCovered"]>
```
seveibar:
```suggestion
  soldermaskWithCopperUnderneath?: Partial<PcbColorMap["soldermaskCovered"]>
```

seveibar (approved):
See the rename, also please make the soldermaskWithCopperUnderneath colors a lighter green, looks at the color SURROUNDING a trace- see how it's darker? That's because there's no copper underneath

image

PR #442, opened 2025-12-06
Full Copper Text Support (Normal + Knockout) in PCB SVG

Abse2001:
Adds full support for `pcb_copper_text`, including normal text rendering, mirroring, rotation, and knockout rectangles with text cutouts. Introduces `opentype.js` for true path extraction and updates snapshot tests.


lib/pcb/svg-object-fns/create-svg-objects-from-pcb-copper-text.ts
```diff
@@ -0,0 +1,351 @@
...
+        const cx = offsetX + Number.parseFloat(qx1) * charScale
+        const cy = offsetY + Number.parseFloat(qy1) * charScale
+        const x = offsetX + Number.parseFloat(qx2) * charScale
+        const y = offsetY + Number.parseFloat(qy2) * charScale
+        return `${cmd2}${cx} ${cy} ${x} ${y}`
+      }
+      return match
+    },
+  )
+}
```
seveibar: needlessly complicated, check the README of @tscircuit/alphabet, it exports lines so that you don't need to parse SVGs

PR #450, opened 2025-12-11
Add board support to component offsets

techmannih:


tests/pcb/pcb-component-board-anchor-offset.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    {
+      type: "pcb_component",
+      pcb_component_id: "comp_2",
+      source_component_id: "source_2",
+      center: { x: -6, y: -4 },
+      width: 3,
+      height: 2.5,
+      layer: "top",
+      rotation: 0,
+      positioned_relative_to_pcb_board_id: "board_1",
```
seveibar: position_mode is not set here

seveibar (approved):

PR #451, opened 2025-12-11
feat: Support NinePointAnchor-based label positioning for voltage probes

Asymtode712:
## PR Description

- Added support for rendering voltage probes with labels at any of 9 anchor positions
- Flipped arrow angle from -50° to -130° for left-aligned labels (center_left, top_left, bottom_left)
- Adjusted text anchor point (end vs start) based on label alignment
- Added comprehensive snapshot test covering all alignment options
- Updated `circuit-json` dependency version to support `label_alignment`

image

### Next Steps

Implementing automatic voltage probe label overlap detection in `core`

Ref: https://github.com/tscircuit/core/issues/1739


seveibar (changes requested):
code is not clean enough, always include unit in names

do not use `isFlipped`- very bad name

you're not handling many arrow positions properly, either error or implement

Asymtode712:
@ShiboSoftwareDev done !
image

seveibar (commented):
top_center, bottom_center, left_center etc. is not properly implemented

in your circuit json, instead of V1 give it the name "TOP_CENTER" etc. so that i can understand the snapshot easily

Asymtode712:
> @seveibar isn't this a bottom-left anchor? Screenshot 2025-12-11 at 9 14 23 PM

image

seveibar (approved):

PR #456, opened 2025-12-13
Fix/panel bounds texture rendering

RaghavArora14:
## Issue
GLTF textures were misaligned because UV mapping and SVG rendering were using different bounds. Providing a single source of truth for texture bounds (board-first, panel fallback) fixes the mismatch and lets downstream consumers use the same box.
## FIx:

- Added getPcbTextureBounds to circuit-to-svg and exported it. It computes board bounds for texture mapping (panel fallback) and throws if neither exists.
- Updated PCB bounds selection in convertCircuitJsonToPcbSvg to track panel bounds separately; allows board-only bounds when requested.
Added tests:
- panel-bounds-calculation.test.ts to cover panel/board viewport selection.
- get-pcb-texture-bounds.test.ts to cover board-first, panel-fallback, and error cases.

cc @ShiboSoftwareDev 

### Snapshots added 
image

shows the viewport uses the full panel bounds, and both boards fit without being clipped, matching the panel/board positions in the test

image

fixes the texture scaling/stretching issue in circuit-json-to-gltf


lib/pcb/utils/compute-bounds.ts
```diff
@@ -0,0 +1,398 @@
...
+    if (pcb_panel_id) {
+      panelBoundsById.set(pcb_panel_id, {
+        minX: centerX - halfWidth,
+        minY: centerY - halfHeight,
+        maxX: centerX + halfWidth,
+        maxY: centerY + halfHeight,
+      })
+    }
+  }
+}
```
ShiboSoftwareDev: see if there is anything in math utils  that would make this code simpler

lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -14,51 +14,58 @@ import {
...
+import { createSvgObjectsFromPcbHole } from "./svg-object-fns/create-svg-objects-from-pcb-hole"
 import { createSvgObjectsFromPcbNoteDimension } from "./svg-object-fns/create-svg-objects-from-pcb-note-dimension"
-import { createSvgObjectsFromPcbNoteText } from "./svg-object-fns/create-svg-objects-from-pcb-note-text"
-import { createSvgObjectsFromPcbNoteRect } from "./svg-object-fns/create-svg-objects-from-pcb-note-rect"
-import { createSvgObjectsFromPcbNotePath } from "./svg-object-fns/create-svg-objects-from-pcb-note-path"
 import { createSvgObjectsFromPcbNoteLine } from "./svg-object-fns/create-svg-objects-from-pcb-note-line"
+import { createSvgObjectsFromPcbNotePath } from "./svg-object-fns/create-svg-objects-from-pcb-note-path"
+import { createSvgObjectsFromPcbNoteRect } from "./svg-object-fns/create-svg-objects-from-pcb-note-rect"
+import { createSvgObjectsFromPcbNoteText } from "./svg-object-fns/create-svg-objects-from-pcb-note-text"
+import { createSvgObjectsFromPcbPanel } from "./svg-object-fns/create-svg-objects-from-pcb-panel"
```
ShiboSoftwareDev: always clean up your diff, don't keep unnecessary changes. important stuff can be missed amidst the noise

lib/pcb/utils/get-board-id.ts
```diff
@@ -0,0 +1,7 @@
+import type { AnyCircuitElement } from "circuit-json"
+import { hasStringProp } from "./has-string-prop"
+
+export const getBoardId = (elm: AnyCircuitElement): string => {
+  if (hasStringProp(elm, "pcb_board_id")) return elm.pcb_board_id
+  return "pcb_board"
```
seveibar: remove this function

lib/pcb/utils/get-empty-bounds.ts
```diff
@@ -0,0 +1,8 @@
+import type { Bounds } from "@tscircuit/math-utils"
+
+export const getEmptyBounds = (): Bounds => ({
+  minX: Number.POSITIVE_INFINITY,
+  minY: Number.POSITIVE_INFINITY,
+  maxX: Number.NEGATIVE_INFINITY,
+  maxY: Number.NEGATIVE_INFINITY,
```
seveibar: this isn't an empty bounds, remove this

lib/pcb/utils/add-rect-to-bounds.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const addRectToBounds = (
+  target: Bounds,
+  center: Point | undefined,
+  width: number | undefined,
+  height: number | undefined,
+): Bounds => {
+  const rect = toRectBounds(center, width, height)
+  return rect ? expandBounds(target, rect) : target
+}
```
seveibar: remove this function

PR #466, opened 2025-12-26
Add SVG rendering support for silkscreen ovals

Abse2001:
Adds full PCB SVG support for pcb_silkscreen_oval elements.

Integrates silkscreen ovals into bounds calculation and layer sorting logic.

Implements SVG ellipse generation with correct transforms, scaling, and rotation handling.

Ensures proper top/bottom silkscreen color mapping and metadata attributes.

Keeps silkscreen behavior consistent with other primitive shapes.

Adds snapshot coverage for silkscreen oval rendering in PCB SVG output.


Abse2001:
@seveibar should we add fill flag to the circuit json or maybe default to fill insted of none?

seveibar (approved):

seveibar:
@Abse2001 wait for users to need more features- we have enough functionality in general now that we shouldn't be adding features speculatively (w.r.t. shape drawing)

PR #473, opened 2026-01-03
Add support for pcb_courtyard_polygon

techmannih:


tests/pcb/pcb-courtyard-polygon.test.ts
```diff
@@ -0,0 +1,42 @@
+import { test, expect } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "lib"
+
+const circuit: any = [
```
ShiboSoftwareDev: please use the correct type

techmannih: oh yes AnyCircuitElement

techmannih: Done!

PR #477, opened 2026-01-09
Improve keepout rendering with diagonal pattern and transparent background

techmannih:
Ref https://github.com/tscircuit/circuit-to-canvas/pull/100/files

### DIFF
--- a/lib/pcb/convert-circuit-json-to-pcb-svg.ts
+++ b/lib/pcb/convert-circuit-json-to-pcb-svg.ts
@@ -305,8 +305,11 @@ export function convertCircuitJsonToPcbSvg(
   // Check if there are any keepouts and add pattern defs if needed
   const hasKeepouts = circuitJson.some((elm) => elm.type === "pcb_keepout")
   if (hasKeepouts) {
-    const keepoutColor = colorMap.keepout ?? "#FF6B6B"
-    children.push(createKeepoutPatternDefs(keepoutColor))
+    children.push(
+      createKeepoutPatternDefs(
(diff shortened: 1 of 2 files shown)


lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -299,6 +302,13 @@ export function convertCircuitJsonToPcbSvg(
     children.push(gridObjects.defs)
   }
 
+  // Check if there are any keepouts and add pattern defs if needed
+  const hasKeepouts = circuitJson.some((elm) => elm.type === "pcb_keepout")
+  if (hasKeepouts) {
+    const keepoutColor = colorMap.keepout ?? "#FF6B6B"
```
imrishabh18: Don't hardcode the default color here, add it to the colorMap and then use it if not present already

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-keepout.ts
```diff
@@ -8,6 +8,115 @@ import {
...
+          height: KEEPOUT_PATTERN_SIZE.toString(),
+          patternUnits: "userSpaceOnUse",
+        },
+        children: createKeepoutPatternLines(keepoutColor),
+      },
+    ],
+  }
+}
+
+function hexToRgba(hex: string, opacity: number): string {
```
imrishabh18: I think we support 8 digits hex too! So, this might not be needed

techmannih: do we need to support in circuit-to-svg?

PR #480, opened 2026-01-15
feat: add support for soldermask_width, soldermask_height and soldermask_center_offset on rectangular smtpads

techmannih:


seveibar (changes requested):
Mae a new test that demonstrates the new functionality. Do not change or break previous tests

tests/pcb/smtpad-soldermask-size-offset.test.ts
```diff
@@ -0,0 +1,115 @@
+import { expect, test } from "bun:test"
```
techmannih: @seveibar I am doing same?

PR #540, opened 2026-04-09
Fix schematic component primitive layering so inner symbols render above box bodies

MustafaMulla29:
`schematic_path` `schematic_circle` primitives attached to a `schematic_component` were being rendered in the global primitive pass, which placed them behind the component’s yellow box fill in SVG output. This made inner symbols appear hidden.

This PR fixes the layering by rendering component-owned schematic primitives inside the component renderers (`with-box`) and `with-symbol` and skipping them in the global pass when the component wrapper is rendered.


seveibar (commented):
the "global pass" concept is too hacky, you can do a more conventional pattern such as a pre-render sort that can make ordering more explicit, this has too many hacks as is

also @MustafaMulla29 this is another inline closure- please stop making inline closures, i think i've mentioned this a couple times

seveibar:
@MustafaMulla29 i meant to say POST-render sort, sorry!

seveibar (changes requested):
const allSvgs = [
   ...schDebugObjectSvgs,
   ...schTraceBaseSvgs,
   // ...
]

const sortedSvgs = allSvgs.sort(customSortFunction)

svgChildren.push(...sortedSvgs)

seveibar:
avoid mutable methods, mutable methods are ONLY used when optimizing and should otherwise be avoided, moveComponentOwnedPrimitivesToComponentGroups is a mutating method, that is also very confusing

seveibar:
@MustafaMulla29 the issue has nothing to do with ordering... you need to not render components that will otherwise be part of a component or something right? Are you trying to render everything within a schematic component into a group?

PR #543, opened 2026-04-10
Wrap every schematic component in a <g> and fix primitive layering

MustafaMulla29:
Previously, `is_box_with_pins === false` components had no `<g>` wrapper, and component-owned primitives (lines, circles, rects, arcs, paths) rendered as top-level SVG elements — causing them to appear below the yellow box. Now every schematic component gets a `<g>`, and owned primitives render inside it after the `box/symbol` so they're visible on top.


lib/sch/convert-circuit-json-to-schematic-svg.ts
```diff
@@ -244,6 +253,11 @@ export function convertCircuitJsonToSchematicSvg(
         }),
       )
     } else if (elm.type === "schematic_line") {
+      if (
+        elm.schematic_component_id &&
+        schematicComponentIds.has(elm.schematic_component_id)
```
seveibar: Just skip if schematic component id is defined no? Why do we need the tracking set?

lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-primitives.ts
```diff
@@ -0,0 +1,84 @@
...
+}: {
+  component: SchematicComponent
+  transform: Matrix
+  circuitJson: AnyCircuitElement[]
+  colorMap: ColorMap
+}): SvgObject[] => {
+  const svgObjects: SvgObject[] = []
+
+  for (const elm of circuitJson) {
+    if (
```
seveibar, line 26: early early continue here

lib/sch/convert-circuit-json-to-schematic-svg.ts
MustafaMulla29: But some tests in this repo has schematic_component_id but not component associated to it @seveibar , so is that wrong? should i remove schematic_component_id from those tests?

MustafaMulla29: Ok then i can do this refactor and remove them from tests

seveibar (approved):

seveibar:
avoid ternaries btw

seveibar:
the only mistake in this code is the ternary means the code will have a messy diff when a third case is added, you should know this when creating it and identify a better structure than a ternary

ternaries are almost always bad

MustafaMulla29:
> the only mistake in this code is the ternary means the code will have a messy diff when a third case is added, you should know this when creating it and identify a better structure than a ternary
> 
> ternaries are almost always bad

Got it! Fixed it

PR #554, opened 2026-05-06
fix(pcb-bounds): tolerate route entries with nested start/end coords (unblocks pcb_trace rendering)

gsdali:
## Summary

`updateTraceBounds` in `getPcbBoundsFromCircuitJson` was calling `distance.parse(point.x)` and `distance.parse(point.y)` directly on each route entry. zod throws `ZodError` when the value is undefined — so any `pcb_trace` containing a route entry without top-level x/y (e.g. `route_type: "through_obstacle"`, which nests its coordinates inside `start` / `end` objects) crashed the entire bounds parser, and therefore the entire renderer, before ANY trace got a chance to draw.

**Symptom:** a circuit with 38 `pcb_trace` records produced 0 trace path elements in the SVG output, even for the 35 traces that contained only `wire` and `via` route entries (no `through_obstacle`). The `through_obstacle` entries in 3 of the 38 traces were enough to crash bounds for all of them.

## Fix

Skip route points without valid x/y, and pick up nested `start.x`/`start.y` and `end.x`/`end.y` so `through_obstacle` entries contribute their actual coordinates to the bounding box (instead of being silently ignored once they're not crashing things):

```ts
function updateTraceBounds(route: Point[]) {
  let updated = false
  const tryUpdate = (px: unknown, py: unknown) => {
    if (px == null || py == null) return
    const x = distance.parse(px)
    const y = distance.parse(py)
    if (x === undefined || y === undefined) return
    minX = Math.min(minX, x); /* ... */
    updated = true
  }
  for (const point of route) {
    tryUpdate((point as any)?.x, (point as any)?.y)
    tryUpdate((point as any)?.start?.x, (point as any)?.start?.y)
    tryUpdate((point as any)?.end?.x, (point as any)?.end?.y)
  }
  if (updated) hasBounds = true
}
```

## Test plan

- [x] All 203 existing tests pass unchanged
- [x] New `tests/pcb/pcb-trace-with-through-obstacle-renders.test.ts`:
   - through_obstacle in a route doesn't throw, emits trace paths
   - clean wire+via route in the same circuit also renders (the crash previously took out sibling traces)
- [x] `bun test`: 205 pass total
- [x] `biome format`: clean

## What this DOESN'T fix (separate follow-up)

`createSvgObjectsFromPcbTrace` itself still accesses `start.x` / `end.x` directly when building SVG path `d` attributes, so SVG segments *adjacent to* a `through_obstacle` entry currently render with `NaN` coordinates. Fixing it requires picking the right nested coord per pair direction (start vs end of the obstacle for incoming vs outgoing segments).

The bounds-crash fix here unblocks rendering for the common case (routes without `through_obstacle` entries) — which is the majority of autorouter output. A follow-up PR should fix the trace renderer itself for the through_obstacle case.

## Repro

A 24 × 22 mm 2-layer ISP3080-UX daughter board with the autorouter producing 38 `pcb_traces`, 3 of which contained `through_obstacle` entries from layer transitions through copper pour. Visually, all 38 traces were missing from `tsci snapshot --pcb-only`. After this fix, 35/38 render correctly (the 3 with through_obstacle still have NaN-coord segments pending the renderer follow-up).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/pcb/get-pcb-bounds-from-circuit-json.ts
+++ b/lib/pcb/get-pcb-bounds-from-circuit-json.ts
@@ -406,15 +406,27 @@ export function getPcbBoundsFromCircuitJson(
 
   function updateTraceBounds(route: Point[]) {
     let updated = false
-    for (const point of route) {
-      const x = distance.parse(point?.x)
-      const y = distance.parse(point?.y)
-      if (x === undefined || y === undefined) continue
+    const tryUpdate = (px: unknown, py: unknown) => {
+      // distance.parse throws on undefined/null; guard before parsing
+      // so a single missing-coords route entry doesn't crash bounds
+      // computation for the entire circuit JSON.
+      if (px == null || py == null) return
+      const x = distance.parse(px)
+      const y = distance.parse(py)
+      if (x === undefined || y === undefined) return
       minX = Math.min(minX, x)
       minY = Math.min(minY, y)
       maxX = Math.max(maxX, x)
       maxY = Math.max(maxY, y)
       updated = true
+    }
+    for (const point of route) {
+      // Most route entries (wire, via) have top-level x/y. Some
+      // entries (e.g. through_obstacle) nest coordinates inside
+      // start/end objects instead — handle both.
+      tryUpdate((point as any)?.x, (point as any)?.y)
+      tryUpdate((point as any)?.start?.x, (point as any)?.start?.y)
+      tryUpdate((point as any)?.end?.x, (point as any)?.end?.y)
     }
     if (updated) {
       hasBounds = true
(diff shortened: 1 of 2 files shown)


gsdali:
Bumping per the stale-bot tag. The single failing CI check (`tests/sch/draw-ports-connected.test.tsx` — "drawPorts option does not render circles for connected ports") is unrelated to this PR's changes; the test exercises schematic port circle rendering with the capacity-mesh autorouter and times out on the autorouter, not on the bounds calc. The same test has timed out on recent `main` runs as well (e.g. workflow run 25071743990). This PR's diff is confined to the bounds-calculation path for `pcb_bounds` route entries with nested `start.x/y, end.x/y` (which `through_obstacle` entries use). #555 stacks on top to do the actual render.

PR #555, opened 2026-05-06
fix(pcb-trace render): use nested start/end coords for through_obstacle entries

gsdali:
## Summary

Follow-up to the bounds-crash fix in #554. With bounds parsing tolerant of nested-coord route entries, the renderer is reached for `through_obstacle` traces — but `createSvgObjectsFromPcbTrace` itself was still reading `start.x` / `end.x` directly when forming SVG path `d` attributes. For pairs adjacent to a `through_obstacle` entry (which has no top-level x/y, only `start: {x,y}` / `end: {x,y}`), this produced `M NaN NaN L NaN NaN` — invalid SVG that browsers silently skip.

This PR teaches the renderer how to read the nested coords correctly.

## Fix

Add helpers `entryCoord(p)` / `exitCoord(p)` that read the appropriate coord per pair direction:

- segment START coord uses the previous entry's "exit"  
  (`entry.end?.{x,y}` for through_obstacle, otherwise `entry.{x,y}`)
- segment END coord uses the next entry's "entry"  
  (`entry.start?.{x,y}` for through_obstacle, otherwise `entry.{x,y}`)

Skip the segment if either coord is missing (defensive — the route data shape guarantees one or the other).

## Test plan

- [x] All 205 tests from #554 still pass unchanged
- [x] New `tests/pcb/pcb-trace-through-obstacle-no-nan.test.ts`:
   - zero-length through_obstacle (start == end) renders without NaN
   - asymmetric through_obstacle (entry coord ≠ exit coord) uses the correct nested coord per pair direction
- [x] `bun test`: 207 pass total
- [x] `biome format`: clean

## Combined effect with #554

Together the two PRs close the visualization gap that caused 38 autorouted `pcb_traces` to render as nothing in `tsci snapshot --pcb-only` output. Verified against an InsightSiP ISP3080-UX daughter board: 300 trace path elements emit, zero NaN coordinates anywhere in the SVG.

## Stacking

This branch is on top of `investigate-pcb-trace-render` (PR #554). Easiest path: merge #554 first, then this. Happy to fold into a single PR if you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/pcb/get-pcb-bounds-from-circuit-json.ts
+++ b/lib/pcb/get-pcb-bounds-from-circuit-json.ts
@@ -406,15 +406,27 @@ export function getPcbBoundsFromCircuitJson(
 
   function updateTraceBounds(route: Point[]) {
     let updated = false
-    for (const point of route) {
-      const x = distance.parse(point?.x)
-      const y = distance.parse(point?.y)
-      if (x === undefined || y === undefined) continue
+    const tryUpdate = (px: unknown, py: unknown) => {
+      // distance.parse throws on undefined/null; guard before parsing
+      // so a single missing-coords route entry doesn't crash bounds
+      // computation for the entire circuit JSON.
+      if (px == null || py == null) return
+      const x = distance.parse(px)
+      const y = distance.parse(py)
+      if (x === undefined || y === undefined) return
       minX = Math.min(minX, x)
       minY = Math.min(minY, y)
       maxX = Math.max(maxX, x)
       maxY = Math.max(maxY, y)
       updated = true
+    }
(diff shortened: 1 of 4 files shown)


gsdali:
Bumping per the stale-bot tag. Same flaky-test note as #554 (this PR is stacked on it; the only failing check is the autorouter-flake `tests/sch/draw-ports-connected.test.tsx`, unrelated to pcb_trace rendering). Together #554 + #555 take the snapshot SVG renderer from drawing 0 `pcb_trace` paths to drawing the full ~300 paths on our 6-layer board.

PR #557, opened 2026-05-06
fix: highlight net labels on trace hover (#1130)

64johnlee:
## Summary
Fixes #1130 - When hovering over a trace, net labels (GND, VCC, etc.) sharing the same source_net_id should also be highlighted.

## Changes
- **create-svg-objects-from-sch-trace.ts**: Add `data-source-net-id` attribute to trace groups and overlays
- **create-svg-objects-for-sch-net-label.ts**: Add `data-source-net-id` attribute to net label SVG paths
- **convert-circuit-json-to-schematic-svg.ts**: 
  - Collect `source_net_id` from traces and net labels
  - Add `buildSourceNetHoverStyles` CSS rules to highlight all traces AND net labels sharing the same `source_net_id` when any trace is hovered

## How it works
When you hover over a trace, the CSS rule inverts colors for all net labels sharing the same `source_net_id`.

## Test
1. Open any schematic with net labels (GND, VCC, etc.) and traces
2. Hover over a trace connected to a net label
3. The net label should now also be highlighted (inverted colors)


64johnlee:
Hi @imrishabh18 - this PR fixes issue #1130 (hover highlighting for net labels). Ready for review. Thanks!

PR #558, opened 2026-05-10
fix: handle through_obstacle route points in updateTraceBounds

tk009999:
The autorouter can produce 'through_obstacle' route points which store coordinates in start/end objects instead of top-level x/y properties. This caused a ZodError crash in distance.parse() when computing PCB bounds for SVG export.

Fix: use safeParse with fallback to start/end coordinates, and also include end point in bounds calculation for through_obstacle segments.

### DIFF
--- a/lib/pcb/get-pcb-bounds-from-circuit-json.ts
+++ b/lib/pcb/get-pcb-bounds-from-circuit-json.ts
@@ -407,14 +407,29 @@ export function getPcbBoundsFromCircuitJson(
   function updateTraceBounds(route: Point[]) {
     let updated = false
     for (const point of route) {
-      const x = distance.parse(point?.x)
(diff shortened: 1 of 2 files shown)


tk009999:
The failing test (drawPorts option does not render circles for connected ports) is a pre-existing timeout in main — the test fixture has a port coordinate mismatch causing the autorouter to hang. Unrelated to this PR.

imrishabh18 (changes requested):
Create a test, I can't verify things work without seeing a visual test snapshot for thi

tk009999:
Added test in tests/pcb/through-obstacle-bounds.test.ts. 203/204 tests pass — the only failure is the pre-existing drawPorts timeout on main (unrelated to this PR).

PR #560, opened 2026-05-10
Handle through-pad PCB trace segments in SVG rendering and bounds

rushabhcodes:
This PR fixes pcb_trace.route handling for the new through_pad route-point variant from circuit-json.

  Changes:

  - normalize PCB trace geometry so through_pad segments are interpreted correctly
  - render through_pad traces across their relevant PCB layers in SVG output
  - include through_pad endpoints in PCB bounds calculation
  - add regression tests covering both bounds and SVG rendering
  - add a snapshot test for a trace that transitions through a pad

  Why this is strong against the guide:

  - this is a bug fix with reproduction in the same PR via tests
  - it touches SVG generation directly
  - it fixes a subtle rendering/data-interpretation issue rather than just types or dependency churn


tests/pcb/through-pad-trace.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        route_type: "through_pad",
+        start: { x: -1, y: 0 },
+        end: { x: 1, y: 0 },
+        width: 0.4,
+        start_layer: "top",
+        end_layer: "bottom",
+      },
+      { route_type: "wire", x: 3, y: 0, width: 0.4, layer: "bottom" },
+    ],
+  },
```
seveibar, line 29: add a pad!

seveibar (approved):

seveibar (approved):

imrishabh18 (approved):

PR #561, opened 2026-05-14
fix knockout text anchor alignment

techmannih:
## Summary

This fixes knockout text placement so `anchor_alignment` is honored for knockout-rendered text in both silkscreen and copper rendering.

## What Changed

- added a shared `getCenteredTextAnchorOffset(...)` helper for `NinePointAnchor`
- applied that offset in knockout silkscreen rendering
- applied that offset in knockout copper rendering
- added regression tests and snapshots for non-center knockout anchors on top and bottom layers

## Root Cause

Knockout text rendering was treating `anchor_position` as if it were always the text center.

That happened because the knockout path generated centered glyph geometry and then rendered it directly at `anchor_position`, but unlike the regular text path it never adjusted for `anchor_alignment`.

As a result:

- `center` anchors rendered correctly
- non-center anchors like `top_left`, `bottom_left`, and `bottom_right` were misplaced
- downstream converters had to compensate upstream by pre-centering the anchor, which is the wrong layer to fix this

## Why Previous Snapshots Did Not Change

Most existing knockout snapshots were already using `anchor_alignment: "center"`.

Since the bug only affected non-center anchors, those snapshots were already exercising the one case that happened to work. That is why the root-cause fix did not materially change the old knockout snapshots.

The new tests cover the missing cases explicitly:

- knockout silkscreen text with `top_left`, `bottom_right`, and bottom-layer `bottom_left`
- knockout copper text with the same non-center alignments

## Impact

- downstream tools can preserve original anchor semantics instead of baking in renderer-specific workarounds
- KiCad-derived knockout text now lines up correctly when the source anchor is not centered
- the fix applies consistently across both silkscreen and copper knockout rendering

## Validation

- `bun test tests/pcb/pcb-silkscreen-text-knockout-anchor-alignment.test.ts tests/pcb/pcb-copper-text-knockout-anchor-alignment.test.ts`
- `bun test tests/pcb/pcb-silkscreen-text-knockout4.test.ts tests/pcb/pcb-copper-text.test.ts`
- validated downstream against the Arduino Nano repro in `kicad-to-circuit-json`


techmannih:
image

tests/pcb/__snapshots__/copper-knockout-anchor-alignment.snap.svg
imrishabh18, line 1: Why is `bottom right` appearing in top layer? Make sure the layer matches with the text

imrishabh18 (changes requested):

tests/pcb/__snapshots__/copper-knockout-anchor-alignment.snap.svg
techmannih: @imrishabh18 Btw the text is represent the anchor_alignment. can you check the anchor_alignment

tests/pcb/__snapshots__/silkscreen-knockout-anchor-alignment.snap.svg
imrishabh18, line 1: Make this simpler, separate out each one of them. I am not sure if I fully get what this means!!!

imrishabh18 (changes requested):

tests/pcb/__snapshots__/silkscreen-knockout-anchor-alignment.snap.svg
techmannih: @imrishabh18 i am fixing the anchor_alignment for knockout text (which does not have any knockout padding). and now i have also marked the anchor position.

techmannih:
@imrishabh18 can you check it

imrishabh18 (approved):
much better and clear, just create one test file with grid layout for all the different anchor position. No need to create separate tests for each of those

PR #562, opened 2026-05-15
Fix PCB text anchor alignment from rendered glyph bounds

techmannih:
## Summary

This fixes the remaining PCB text anchor alignment offset in `circuit-to-svg` by deriving anchor placement from the rendered alphabet path geometry instead of from coarse width/height box math.

## What was wrong

Copper text and knockout text were being anchored from centered layout dimensions rather than the actual rendered glyph bounds. That worked for some alignments, but `top_*`, `bottom_*`, and side-aligned cases could still land slightly off because the alignment math was not using the true outline extents.

During the first pass of the fix, regular silkscreen text was also moved onto the alphabet stroke renderer. That improved consistency, but it changed the visible letterforms, which was a regression.

## What changed

- added a shared helper for PCB alphabet text geometry
- compute text bounds from the rendered path segments themselves
- derive anchor offsets from those real bounds for copper text and knockout text
- use the shared geometry in both copper text and silkscreen knockout rendering
- remove the older centered width/height anchor helper
- keep regular silkscreen text on the existing SVG `<text>` renderer so its visual design stays unchanged
- refresh affected snapshots to match the corrected placement

## Why this approach

This keeps the fix in the rendering model instead of layering more special-case offsets on top:

- one source of truth for alphabet-path text bounds
- anchor placement follows the real rendered shape
- knockout rectangles inherit the same geometry and stay aligned with the text they mask
- regular silkscreen text keeps its original appearance

## Impact

- fixes residual side/top/bottom anchor drift for PCB copper text and knockout text
- keeps silkscreen text visuals stable
- makes future PCB text alignment changes easier to reason about because the geometry and anchoring logic are centralized

## Validation

```sh
bun test --timeout 30000
```

Passed: `207 pass, 0 fail`


tests/pcb/__snapshots__/silkscreen-knockout-anchor-alignment.snap.svg
imrishabh18, line 1: This looks correct!

imrishabh18 (approved):
This looks correct from the test, but I am not sure about the code changes, cause the diff is huge and difficult to review. It seems like you have also refactor some of the files.

Take a review from Seve too

seveibar (approved):
good, though i think the knockout alignment is using the wrong anchor (it's using an inner text anchor by default rather than the outer part)

image

techmannih:
Yes exactly. I can fix it.

PR #563, opened 2026-05-21
Render route vias from trace points

techmannih:
## What changed
Render visible PCB via geometry for `pcb_trace.route` entries with `route_type: "via"` when the circuit JSON does not already include a matching top-level `pcb_via` element.

Also remove degenerate zero-length trace segments that were being emitted around via waypoints.

## Why
The renderer already used route-level via points to switch segment layers, but only standalone `pcb_via` elements produced visible via circles and drill holes. When a circuit used route-only vias, the SVG showed layer changes without any via geometry.

## Impact
PCB snapshots now show vias for traces that encode them directly in the route.

This keeps existing explicit `pcb_via` elements authoritative and avoids double-drawing when both representations are present.

## Root cause
`route_type: "via"` was handled during trace segmentation, but not during SVG object creation. That left behind zero-length trace segments at via coordinates and no actual via render objects.

## Validation
- `bun test tests/pcb`
- `bun test tests/pcb/bottom-trace-in-blue.test.ts tests/pcb/bottom-top-trace.test.ts tests/pcb/multi-layer-traces.test.ts tests/pcb/through-pad-trace.test.ts tests/pcb/four-layer-routing-circuit.test.ts`

## Notes
Route-level via points do not currently carry explicit diameter fields in the schema, so the renderer uses a conservative fallback size for route-only vias unless a matching `pcb_via` element is present.


lib/pcb/svg-object-fns/create-svg-objects-from-pcb-trace.ts
```diff
@@ -1,9 +1,13 @@
...
+import { distance, type PCBVia, type PcbTrace } from "circuit-json"
 import type { INode as SvgObject } from "svgson"
 import { applyToPoint } from "transformation-matrix"
 import { layerNameToColor } from "../layer-name-to-color"
 import type { PcbContext } from "../convert-circuit-json-to-pcb-svg"
 import { getPcbTraceSegments } from "../get-pcb-trace-segments"
+import { createSvgObjectsFromPcbVia } from "./create-svg-objects-from-pcb-via"
+
+const DEFAULT_ROUTE_VIA_OUTER_DIAMETER = 0.6
+const DEFAULT_ROUTE_VIA_HOLE_DIAMETER = 0.3
```
seveibar: this shouldn't be encoded here- use the pcb_board min via sizes

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-trace.ts
```diff
@@ -84,5 +88,103 @@ export function createSvgObjectsFromPcbTrace(
     }
   }
 
+  for (const [index, point] of trace.route.entries()) {
+    if (point.route_type !== "via") continue
+    if (hasStandaloneViaAtRoutePoint(trace, point, ctx)) continue
```
seveibar: this shouldn't be needed, remove? What happens if you remove this

seveibar (approved):
see important comments

seveibar:
please get another approval for me or rishabh before merge to make sure fixes get implemented correctly

lib/pcb/svg-object-fns/create-svg-objects-from-pcb-trace.ts
```diff
@@ -84,5 +92,131 @@ export function createSvgObjectsFromPcbTrace(
...
+    board?.min_via_pad_diameter,
+  )
+
+  // Older circuit-json payloads can omit board-level via DRC fields.
+  const fallbackHoleDiameter = Math.max(adjacentTraceWidth, 0.3)
+  const holeDiameter = boardMinViaHoleDiameter ?? fallbackHoleDiameter
+  const outerDiameter = Math.max(
+    boardMinViaPadDiameter ?? Math.max(fallbackHoleDiameter * 2, 0.6),
+    holeDiameter * 2,
+  )
```
imrishabh18: Why max? Outer diameter should be the `min_via_pad_diameter`

lib/pcb/get-pcb-trace-segments.ts
```diff
@@ -76,3 +83,21 @@ function isInsideCopperPour(point: PcbTraceRoutePoint): boolean {
...
+  const by = distance.parse(b.y)
+
+  if (
+    ax === undefined ||
+    ay === undefined ||
+    bx === undefined ||
+    by === undefined
+  ) {
+    return a.x === b.x && a.y === b.y
+  }
```
imrishabh18: Why are you strict equality operator for `undefined`?

imrishabh18: What's the point of this

imrishabh18 (approved):

techmannih:
image

PR #564, opened 2026-05-22
fix knockout text anchor alignment

techmannih:
## Summary
Fix knockout text alignment so it anchors against the outer padded knockout bounds instead of the inner glyph bounds.

## Why
Knockout text was using the inner text geometry as its default anchor reference. That made alignments like left/right/top/bottom land on the glyph strokes instead of the full knockout area.

## What changed
- exposed the shared bounds-based anchor offset helper for reuse
- applied knockout alignment using the padded outer knockout bounds in both silkscreen and copper text renderers
- kept non-knockout text behavior unchanged
- updated the knockout alignment snapshots to reflect the corrected positioning

## Impact
Knockout text now aligns the same way users expect other anchored primitives to align, especially when padding is present.

## Root cause
The knockout rect was padded after the text geometry had already been anchored, so the anchor reference effectively came from the inner text bounds rather than the outer knockout box.

## Validation
- `bun test tests/pcb/pcb-silkscreen-text-knockout-anchor-alignment.test.ts tests/pcb/pcb-copper-text-knockout-anchor-alignment.test.ts`
- `bun test tests/pcb/pcb-silkscreen-text-knockout1.test.ts tests/pcb/pcb-silkscreen-text-knockout2.test.ts tests/pcb/pcb-silkscreen-text-knockout3.test.ts tests/pcb/pcb-silkscreen-text-knockout4.test.ts tests/pcb/pcb-silkscreen-text-knockout5.test.ts tests/pcb/pcb-silkscreen-text-knockout-anchor-alignment.test.ts tests/pcb/pcb-copper-text-knockout-anchor-alignment.test.ts tests/pcb/pcb-copper-text.test.ts tests/pcb/pcb-copper-text-core.test.tsx`


seveibar (approved):
excellent

PR #565, opened 2026-05-27
Add board soldermask and silkscreen colors (no merge) 

mohan-bee:
Adds support for `pcb_board.solder_mask_color` and `pcb_board.silkscreen_color` in PCB SVG realistic rendering.

These board colors are only used when `showSolderMask` is enabled, matching the realistic render path used for board texture generation. Existing default/debug colors remain unchanged outside soldermask rendering, and explicit `colorOverrides` still take priority.


seveibar (changes requested):
Cc @Abse2001 i dont think this is right- we only use certain colors when we’re rendering in realistic mode no?

imrishabh18 (approved):
Take review from Seve

mohan-bee:
idk what happened this process update is taking too long and not allowing any commits, : 
image


i have added the change in new PR : https://github.com/tscircuit/circuit-to-svg/pull/566

PR #566, opened 2026-05-27
Add board soldermask and silkscreen colors

mohan-bee:
https://github.com/tscircuit/circuit-to-svg/pull/565
 
 Adds support for pcb_board.solder_mask_color and pcb_board.silkscreen_color in PCB SVG realistic rendering.

These board colors are only used when showSolderMask is enabled, matching the realistic render path used for board texture generation. Existing default/debug colors remain unchanged outside soldermask rendering, and explicit colorOverrides still take priority.

increased the timeout and format up  here net-label.test.tsx


seveibar:
@mohan-bee  this looks bad to me- like you're changing the color overrides system. Can you get @Abse2001 to help you with this one- i think he understands how to apply this

also it would help to use a more realistic board. Right now I don't see why we would ever want the output you're showing

You might also want to start in circuit-to-canvas since that might be used by the 3d viewer. You have to keep in mind this color SHOULD NOT APPLY in most cases, the only time it applies is when we want to realistically render the board- otherwise the colors are standard

PR #577, opened 2026-06-18
Fix literal text rendering in schematic symbols

mohan-bee:


AnasSarkiz (commented):
Doesn't look same as the https://symbols.tscircuit.com ?

Image

mohan-bee:
> Doesn't look same as the https://symbols.tscircuit.com ?
> 
> Image

maybe i can increase the fontsize ?  will that work ?

AnasSarkiz:
@mohan-bee, we just need it to be as the voltmeter, the voltmeter draws the V shape using normal SVG/path primitives

mohan-bee:
> @mohan-bee, we just need it to be as the voltmeter, the voltmeter draws the V shape using normal SVG/path primitives

@AnasSarkiz   actually in the schematic symbol the voltmeter is defined as path primitives but in case of ammeter it is not, it is just defined as text or we have to expose the text as path from schematic symbols maybe this might work. 

but still why would we want a text to be drawn as path ?

mohan-bee:
or else we can expose that as path from schematic symbols ?

ShiboSoftwareDev:
doesn't look symmetric

ShiboSoftwareDev:
I think you need to change how it's drawn in the symbols repo

mohan-bee:
> I think you need to change how it's drawn in the symbols repo

yes that's what i am thinking too , the schematic symbols now exposes that as text, can we make that expose as path ??

mohan-bee:
fixed

PR #580, opened 2026-06-21
Add scope-style channel legend for simulation display options

ShiboSoftwareDev:
image

image

ref:
image


AnasSarkiz (commented):
I think when all are showing voltage, we need to make the y-axis show volts. And could u add a test that has current and voltage in the same graph?

tests/sim/__snapshots__/probe-display-options.snap.svg
AnasSarkiz, line 1: L1 shows current? because its using 0 V

ShiboSoftwareDev:
@AnasSarkiz  We can’t show volts on the shared y-axis because each scope channel can use its own center,
  offset, and volts-per-div scale, so the same y-position can represent different voltages for
  different channels.

AnasSarkiz (approved):

tests/sim/simulation-graph.test.ts
```diff
@@ -393,6 +401,143 @@ test("uses display divisions for mixed voltage and current graphs with display o
...
+
+    const probeId = `simulation_voltage_probe_ch${channelNumber}`
+    circuit.push(
+      {
+        type: "simulation_voltage_probe",
+        simulation_voltage_probe_id: probeId,
+        name: `CHANNEL_${channelNumber}`,
+        color,
+        display_options: {
+          label: `CH${channelNumber}_SIG`,
```
seveibar: Bad spec, change to display_name to match convention

tests/sim/simulation-graph.test.ts
```diff
@@ -393,6 +401,143 @@ test("uses display divisions for mixed voltage and current graphs with display o
...
+          center: channelNumber,
```
seveibar: I dont know what this is (bad spec)

tests/sim/simulation-graph.test.ts
```diff
@@ -393,6 +401,143 @@ test("uses display divisions for mixed voltage and current graphs with display o
...
+          offset_divs: channelNumber - 3,
```
seveibar: Bad spec (what is this?)

tests/sim/simulation-graph.test.ts
```diff
@@ -393,6 +401,143 @@ test("uses display divisions for mixed voltage and current graphs with display o
...
+          units_per_div: channelNumber === 1 ? 0.05 : channelNumber,
```
seveibar: Again what is this for a probe? Bad spec

seveibar (commented):
Bad spec, nesting is usually a mistake and on this case you have legend properties unclear labeled as "display_options"

The actual output is ok- but the spec is not good- very unclear

tests/sim/simulation-graph.test.ts
seveibar: Why do we even need this? Name is probably sufficient

lib/sim/convert-circuit-json-to-simulation-graph-svg.ts
```diff
@@ -903,6 +918,189 @@ function createLegend(
...
+
+function formatScopeNumber(value: number): string {
+  if (!Number.isFinite(value)) return "0"
+
+  const absValue = Math.abs(value)
+  const maximumFractionDigits =
+    absValue >= 100 ? 0 : absValue >= 10 ? 1 : absValue >= 1 ? 2 : 3
+
+  return Number.parseFloat(value.toFixed(maximumFractionDigits)).toString()
+}
```
seveibar: 1000+ lines and youre adding more functions, clean this up!!!!!! Do not make a bad problem worse!

tests/sim/simulation-graph.test.ts
seveibar: why are we saying “units”? We know its volts!!!

seveibar (changes requested):
These graphs are wrong! The units per div (VOLTS per div) are incorrectly marked on the left axis!

seveibar:
With a scope graph there is an axis per probe colored for that probe

seveibar (approved):

seveibar:
What’s the reference for the multi axis approach for the graph?

PR #582, opened 2026-06-27
Add support for schematic_sheet, the subcircuit_id linked to the schematic_sheet is rendered in the center of the sheet

imrishabh18:


seveibar (approved):
Hmm why is the resistor so big? Maybe try to keep it meaningful, use the equivalent of an A7 sheet soze (or whwtever kicad does but scaled for our 1mm resistors…)

seveibar:
@imrishabh18 i think we should revert this- it is built on a misunderstanding of how schematic_sheet works. Unfortunately it is not reverting automatically!

lib/sch/convert-circuit-json-to-schematic-svg.ts
```diff
@@ -1,4 +1,4 @@
-import type { AnyCircuitElement } from "circuit-json"
+import type { AnyCircuitElement, SchematicSheet } from "circuit-json"
```
seveibar, line 1: this file should not have been edited!!!!! A major anti-pattern with an inline closure was introduced!!!!

PR #585, opened 2026-06-28
Schematic Sheet implementation w/ filtering

seveibar:


lib/sch/schematic-sheet-utils.ts
```diff
@@ -1,11 +1,15 @@
...
-export const DEFAULT_SCHEMATIC_SHEET_HEIGHT = 210
+const KICAD_RESISTOR_PIN_SPAN_MM = 10.16
+const TSCIRCUIT_RESISTOR_PIN_SPAN = 1.1
+export const SCHEMATIC_UNIT_TO_MM =
+  KICAD_RESISTOR_PIN_SPAN_MM / TSCIRCUIT_RESISTOR_PIN_SPAN
 
-export const SCHEMATIC_SHEET_GAP = 20
-export const SCHEMATIC_SHEET_INNER_MARGIN = 5
+export const DEFAULT_SCHEMATIC_SHEET_WIDTH = 297 / SCHEMATIC_UNIT_TO_MM
+export const DEFAULT_SCHEMATIC_SHEET_HEIGHT = 210 / SCHEMATIC_UNIT_TO_MM
```
seveibar, line 9: @imrishabh18 as a temporary hack I've computed the effective A4 size given our scaling

this is just building on the scaling issue further- users may want to specify other standard paper sizes for the schematic etc. we need to resize all the symbols in the schematic-symbols project most likely to fit the conventional proportions

our original scaling was "a resistor is 1" so our "unit" is "resistorlengths". Conveniently, this means that pin spacing comes our to 0.2 resistor lengths. So the numbers are pretty easy to keep track of.

In kicad-land. A resistor is 400mil and pin spacing is 100mil. This translates horribly to mm and sort of forces you to size things according to the imperial system (100mil = 2.54mm, 400mil = 10.16mm).

We may be able to continue using "our unit" by giving it an alias, e.g. "1rl" could be "1 resistor length". This would help with the migration of code but switching to 1 = "400mil" is almost as easy and is more conventional

CC @MustafaMulla29 this would resolve the crazy scaling hacks you had to do to get symbol kicad conversion working decently. CC @Sang-it because symbol stuff

seveibar: i have no idea when we would want to do this because it's going to cause so many issues and slow down other efforts

PR #586, opened 2026-06-29
feat(sch): add convertCircuitJsonToStackedSchematicSheetsSvg for multi-sheet schematics

MustafaMulla29:
`convertCircuitJsonToSchematicSvg` only renders one schematic sheet at a time. This adds `convertCircuitJsonToStackedSchematicSheetsSvg`, which renders **every** sheet stacked vertically into a single SVG — one labeled panel per sheet — so you can eyeball all sheets at once. Primarily a debugging aid for multi-sheet designs.

## Behavior
- Sheets are ordered by `sheet_index` ascending (top → bottom), full width.
- Each panel is the regular single-sheet render of that sheet (sheet border included), so panels look identical to the normal schematic view.
- Each panel shows the sheet's `display_name` (falling back to `name`, then `Sheet N`) in a band above the frame, left-aligned to the frame's top-left corner.
- Panel height defaults to the A4 sheet aspect ratio (297:210) when `height` is omitted, so panels fill the width without letterboxing.
- Falls back to `convertCircuitJsonToSchematicSvg` when there are fewer than 2 sheets.

## Implementation notes

- Follows the existing composite-SVG pattern from `convertCircuitJsonToSchematicSimulationSvg` (render sub-SVGs → `parseSync` → compose).
- Extracted the shared composition helpers (`ensureElementNode`, `cloneSvgObject`, `translateNestedSvg`, `formatNumber`) into `lib/utils/svg-object-utils.ts` and refactored the simulation converter to use them. The simulation refactor is behavior-identical (its snapshot is unchanged).


lib/sch/convert-circuit-json-to-stacked-schematic-svg.ts
```diff
@@ -0,0 +1,204 @@
...
+ * single SVG, one panel per sheet with the sheet's display name on top. This is
+ * primarily a debugging aid - `convertCircuitJsonToSchematicSvg` only renders a
+ * single sheet at a time, so this makes it easy to eyeball every sheet at once.
+ *
+ * Each panel is the regular single-sheet render of that sheet (sheet border
+ * included), so the panels look exactly like the normal schematic view. When
+ * fewer than two sheets are present this falls back to the single-sheet
+ * renderer.
+ */
+export function convertCircuitJsonToStackedSchematicSvg(
```
imrishabh18: Don't couple the `schematic_sheet` with `convertCircuitJsonToStackedSchematicSvg`, the name doesn't say that this is for sheet's

imrishabh18 (changes requested):

imrishabh18 (approved):

PR #592, opened 2026-07-01
fix: size legend columns to available width instead of a fixed 3-column cap

rushabhcodes:
## Summary

The simulation oscilloscope legend was hard-coded to a maximum of **3 columns** (`SCOPE_LEGEND_MAX_COLUMNS = 3`), regardless of how wide the rendered SVG actually was. On scopes with 4+ trace channels this forced the channel cards to wrap into an unbalanced extra row even when there was ample horizontal space to lay them out on fewer rows — wasting vertical space and pushing the legend further below the plot than necessary.

This PR makes the legend column count **derive from the available width**, so the legend packs as many fixed-width channel cards per row as physically fit.

## Problem (before)

`getScopeLegendGridLayout()` ignored `containerWidth` for the column count and capped columns at 3:

- **4 channels** → forced into `3 + 1` (two rows, one orphaned card)
- **5 channels** → forced into `3 + 2` (two rows)

…even when the SVG was wide enough to fit all cards on fewer rows.

## Fix (after)

Columns are now computed from the container width using the real card geometry (`CARD_WIDTH = 118`, `CARD_GAP = 18`):

```ts
const columnsThatFit = Math.floor(
  (containerWidth + SCOPE_LEGEND_CARD_GAP) /
    (SCOPE_LEGEND_CARD_WIDTH + SCOPE_LEGEND_CARD_GAP),
)
const columns = Math.max(1, Math.min(columnsThatFit, graphCount))
```

For the standard `width: 500` scope render:

- **4 channels** → single row of `4` (was `3 + 1`)
- **5 channels** → `4 + 1` (was `3 + 2`)

The legend stays centered within the graph's viewBox — no overflow — and is clamped to a minimum of one column so narrow containers degrade gracefully.

## Reproduction & test

Added `tests/sim/simulation-graph-four-channel-example.test.tsx`: a placeholder resistor schematic plus **four oscilloscope trace channels** (VO / L1 / L2 / IL). It exercises the exact >3-channel case that previously wrapped, asserts the legend renders all four channels, and locks the corrected single-row layout in with an SVG snapshot.

## Changes

- `lib/sim/simulation-graph-svg/create-legend/get-scope-legend-grid-layout.ts` — replace the fixed `SCOPE_LEGEND_MAX_COLUMNS = 3` cap with a width-based fit.
- `tests/sim/simulation-graph-four-channel-example.test.tsx` — new four-channel scope reproduction + snapshot.
- Updated `four-channel-scope-legend` / `five-channel-scope-legend` snapshots to reflect the tighter, correctly-sized layout.

## Testing

- `bun test tests/sim/` → 17 pass / 0 fail
- `npx tsc --noEmit` → clean
- `biome format` → clean


tests/sim/__snapshots__/simulation-graph-tps63802-example.snap.svg
ShiboSoftwareDev, line 1: your test is not a pfm buck operation, just call it a "four channel..." and instead of that sch just put a placeholder resistor.

ShiboSoftwareDev (approved):
lg, just change the sch and title for the test and we can merge

PR #594, opened 2026-07-03
add pad name data attributes

seveibar:


seveibar:
CC @Abse2001 @imrishabh18  maybe will help for debugging (for AI and humans with inspect element)

Issue #595, opened 2026-07-05
getPcbBoundsFromCircuitJson ignores ccw_rotation for pcb_silkscreen_rect


abdalraof-albarbar:
`getPcbBoundsFromCircuitJson()` measures a `pcb_silkscreen_rect` as an axis-aligned `width` × `height` box and ignores its `ccw_rotation`. But the renderer *does* rotate the rect (`create-svg-objects-from-pcb-silkscreen-rect.ts` emits `transform="translate(...) rotate(-ccw_rotation)"`), so for a rotated rect the exported bounds are too small on the rotated axis and the shape gets clipped by the svg viewport.

Repro — a `7.778 × 1` rect rotated 90°:

```ts
getPcbBoundsFromCircuitJson([{
  type: "pcb_silkscreen_rect", pcb_silkscreen_rect_id: "r1", pcb_component_id: "c1",
  layer: "top", center: { x: 0, y: 0 }, width: 7.778, height: 1,
  stroke_width: 0.1, ccw_rotation: 90,
}])
```

```
actual:   minX=-3.889 maxX=3.889  minY=-0.5   maxY=0.5     (un-rotated box)
expected: minX=-0.5   maxX=0.5    minY=-3.889 maxY=3.889   (rotated AABB)
```

On the default 800×600 canvas the rotated rect renders spanning y ≈ [-18, 618], i.e. clipped top and bottom.

Note: `pcb_smtpad` with `shape: "rotated_rect"` has the same class of issue (its renderer rotates, but the smtpad bounds branch is axis-aligned). Leaving that out here to keep the change small and avoid churning many footprint snapshots — can follow up separately.

abdalraof-albarbar:
Not a bug — @seveibar clarified silkscreen is not used as part of pcb bounds, so a rotated pcb_silkscreen_rect being clipped by the viewport is intended. Closing.

PR #596, opened 2026-07-05
Add failing repro for rotated pcb_silkscreen_rect bounds

abdalraof-albarbar:
Failing repro for #595.

`getPcbBoundsFromCircuitJson` ignores `ccw_rotation` for `pcb_silkscreen_rect`, so a wide rotated rect gets bounds that are too small on the rotated axis and the shape is clipped by the svg viewport. The snapshot here shows the clipped render (the rect runs off the top/bottom).

The bounds assertion is `test.failing` so this stays green; the fix (#597) flips it to `test` and regenerates the snapshot to the un-clipped render.


abdalraof-albarbar:
Closing along with #597 — per @seveibar silkscreen isn't part of pcb bounds, so a clipped rotated silkscreen rect is intended behavior, not a bug.

PR #597, opened 2026-07-05
Account for ccw_rotation in pcb_silkscreen_rect bounds

### LINKED ISSUE #595: getPcbBoundsFromCircuitJson ignores ccw_rotation for pcb_silkscreen_rect

abdalraof-albarbar:
Fixes #595. Builds on the repro in #596.

`getPcbBoundsFromCircuitJson` measured a `pcb_silkscreen_rect` as an axis-aligned `width` × `height` box and ignored `ccw_rotation`, even though the renderer rotates the rect. Wide rotated rects ended up with bounds too small on the rotated axis and were clipped by the svg viewport.

`updateBounds` now takes an optional `ccwRotationDegrees` and, when set, expands the half-extents to the rotated axis-aligned box:

```ts
const angleRadians = (ccwRotationDegrees * Math.PI) / 180
const rotatedHalfWidth = halfWidth * absCos + halfHeight * absSin
const rotatedHalfHeight = halfWidth * absSin + halfHeight * absCos
```

The `pcb_silkscreen_rect` branch passes its `ccw_rotation`; it's a no-op for every other caller. The repro's bounds test flips from `test.failing` to `test`, and its snapshot regenerates — the rect is no longer clipped (before/after visible in the diff).

Scoped to `pcb_silkscreen_rect` (the reported case) to keep the diff small. `pcb_smtpad` `rotated_rect` shares the same class of bug but would churn many footprint snapshots, so I left it for a follow-up — the new `ccwRotationDegrees` param makes that a one-liner later.


seveibar (changes requested):
we don't use silkscreen or fabrication note text for pcb bounds- this is essentially a larger issue in the naming

we need to REMOVE getPcbBounds and change to getCopperPcbBounds and getComprehensivePcbBounds- the concept of "PCB Bounds" often creates confusion for autolayout- the silkscreen is not used when doing autolayout

You CAN create getComprehensivePcbBounds and begin this refactor if you'd like

abdalraof-albarbar:
Reopening — after your follow-up this makes sense as comprehensive bounds, not copper bounds.

To your question "are they currently being used?": yes. `getPcbBoundsFromCircuitJson` folds silkscreen in via `updateSilkscreenBounds`, and `getViewportBounds` uses those full bounds whenever `drawPaddingOutsideBoard` is true (the default) or there's no board — so silkscreen does affect the rendered viewport today.

So rather than stripping silkscreen out, I've kept this PR as-is: it maintains the current (comprehensive) behavior and just makes the silkscreen contribution rotation-correct — a rotated rect now reports its true extent instead of the un-rotated box.

Happy to follow up with the rename to `getComprehensivePcbBounds` + a deprecation alias on the old name in a separate PR after this one merges, so the naming stops implying "copper bounds."

PR #600, opened 2026-07-06
Fix Net Label Heights

seveibar:
fixes this

image

image


lib/sch/svg-object-fns/create-svg-objects-for-sch-net-label.ts
```diff
@@ -49,6 +50,7 @@ export const createSvgObjectsForSchNetLabel = ({
 
   const fontSizePx = getSchScreenFontSize(realToScreenTransform, "net_label")
   const fontSizeMm = getSchMmFontSize("net_label")
+  const halfHeightFsr = NET_LABEL_HEIGHT_MM / fontSizeMm / 2
   const textWidthFSR = estimateTextWidth(labelText || "")
```
Sang-it, line 54: Fsr vs FSR, two different naming convention used, probably use FSR as textWidth uses textWidthFSR.

Sang-it (approved):

lib/sch/svg-object-fns/create-svg-objects-for-sch-net-label.ts
seveibar: we use google-style naming convention: https://github.com/tscircuit/handbook/blob/main/guides/code.md#4-casing

the original code uses the incorrect convention

PR #601, opened 2026-07-07
Add net-label hover highlighting for schematic net labels

MustafaMulla29:
Schematics already highlight a whole net when you hover any of its traces (color inverts across every wire sharing a `subcircuit_connectivity_map_key`). Net labels were left out. This makes real (non-symbol) net labels participate
in the exact same net-hover mechanism:

- Hovering a net label highlights the label **and** the net's traces.
- Hovering a trace now also highlights the net's label(s).
- When a connection is drawn as labels only (e.g. a net shown at each endpoint
  with no wire between them), hovering one label highlights its twin.

Power/ground net labels (rendered as symbols) are intentionally excluded — only real net labels are affected.

## How

Net labels don't carry a connectivity key directly, so it's resolved from `source_net_id`, which links to the same key the net's traces use:

- **Named nets** (`<trace from="net.MYNET" />`) reference a `source_net`
  element carrying `subcircuit_connectivity_map_key`.
- **Named/auto-named traces** (`<trace name="R1_C1" />` or unnamed) have no
  `source_net` element — their `source_net_id` *is* that connectivity key.

The new `getSchematicNetLabelConnectivityKey` helper prefers the `source_net` lookup and falls back to the raw `source_net_id`, covering both. The fallback degrades safely: it can only ever group elements of the *same* net.

The renderer now wraps each real net label's outline + text in a `<g class="net-label sch-net-label" data-subcircuit-connectivity-map-key=…>` — structurally identical to how components (`sch-component`) and traces (`trace sch-trace`) are already grouped. `buildNetHoverStyles` is extended so net-label groups are both a hover trigger and a highlight target alongside the net's traces, plus a per-label `:hover` fallback (mirroring the existing per-trace fallback)

**Linking into schematic-viewer**
Screenshot 2026-07-07 at 12 32 07 PM


lib/utils/get-schematic-net-label-connectivity-key.ts
```diff
@@ -0,0 +1,28 @@
...
+ * we prefer the `source_net` lookup and fall back to the raw `source_net_id`.
+ */
+export function getSchematicNetLabelConnectivityKey(
+  schNetLabel: SchematicNetLabel,
+  circuitJson: AnyCircuitElement[],
+): string | undefined {
+  const sourceNetId = schNetLabel.source_net_id
+  if (!sourceNetId) return undefined
+  const sourceNet = su(circuitJson).source_net.get(sourceNetId)
+  return sourceNet?.subcircuit_connectivity_map_key ?? sourceNetId
```
imrishabh18: This seems wrong, how can you just return the `source_net_id` when you can't find the `subcircuit_connectivity_map_key`. Both of them are different

imrishabh18 (changes requested):

seveibar (changes requested):
Doesnt highlight the traxes.

also i would consider moving the implementation for anything interactive into schematic-viewer- it might have been a mistake to include it in the base svg. But your are welcome to try provided you do the highlighting directly

that said, the bright pink looks cheap and was mostly added for debugging- the convention which is impossible-ish to do in the base svg is to fade all other nets and unrelated chips when hovering!!!!

MustafaMulla29:
> Doesnt highlight the traxes.
> 
> also i would consider moving the implementation for anything interactive into schematic-viewer- it might have been a mistake to include it in the base svg. But your are welcome to try provided you do the highlighting directly
> 
> that said, the bright pink looks cheap and was mostly added for debugging- the convention which is impossible-ish to do in the base svg is to fade all other nets and unrelated chips when hovering!!!!

@seveibar I tried linking circuit-to-svg in schematic-viewer to test the hovering and it was working.

seveibar:
My point is that the right design is only available in js, brcayse we should be fading unrelated chips and nets on hover. The base svg does not need interaction

MustafaMulla29:
> My point is that the right design is only available in js, brcayse we should be fading unrelated chips and nets on hover. The base svg does not need interaction

Ok sg!

PR #602, opened 2026-07-07
Tag schematic net label elements with data-schematic-net-label-id and remove hover CSS from base SVG

MustafaMulla29:


seveibar (commented):
why are you wrapping in a group

MustafaMulla29:
> why are you wrapping in a group

So in schematic-viewer it will be easy to find right?(using id)

seveibar (approved):

PR #605, opened 2026-07-08
Count all rendered elements in getComprehensivePcbBounds

### LINKED ISSUE #603: getComprehensivePcbBounds drops several rendered elements (copper clipped from viewport)
`getComprehensivePcbBounds` (formerly `getPcbBoundsFromCircuitJson`) is used to size the exported SVG viewport, but it ignores the extent of several element types that the renderer *does* draw — so their copper/silkscreen gets clipped, and in the degenerate case (via/plated-hole alone) the viewport collapses to a zero-size point.

Missing / wrong:
- **pcb_via** — no branch; `outer_diameter` ignored (falls into the generic zero-size point fallback)
- **pcb_plated_hole** — no branch at all; every shape's pad extent ignored (edge connectors, castellated/mounting pads clip)
- **pcb_smtpad `rotated_rect`** — measured as an un-rotated width/height box
- **pcb_smtpad `rotated_pill`** — dropped entirely (not in the shape list)
- **pcb_copper_pour `brep`** — rect/polygon handled, brep dropped
- **pcb_silkscreen_pill** — rendered but missing from the silkscreen handling

```ts
getComprehensivePcbBounds([
  { type: "pcb_via", pcb_via_id: "v", x: 0, y: 0, outer_diameter: 8, hole_diameter: 4,
    layers: ["top","bottom"], from_layer: "top", to_layer: "bottom" },
])
// bounds -> {0,0,0,0}  (should be +/-4)
```

All six are the same root cause: the bounds function isn't exhaustive vs the per-element renderers.

abdalraof-albarbar:
Fixes #603. Builds on the repros in #604.

`getComprehensivePcbBounds` now includes the extent of every element the renderer draws:

- **pcb_plated_hole** — dispatch per shape (circle/oval/pill, rect-pad and rotated rect-pad, polygon pad) to the outer pad extent
- **pcb_via** — `outer_diameter`
- **pcb_smtpad** `rotated_rect` / `rotated_pill` — expand to the rotated axis-aligned box via `ccwRotationDegrees`
- **pcb_copper_pour** `brep` — outer-ring vertices; and rect pours now honor their `rotation`
- **pcb_silkscreen_pill** — center + width/height

Before → after (viewport now contains the full copper/silkscreen instead of clipping it or collapsing to a point):

before/after grid

- 7 numeric repros flip from `test.failing` to `test`; each of the 5 element snapshots regenerates to the un-clipped render (diffs visible in this PR)
- 5 existing snapshots (rotated-pad / plated-hole / error scenes) regenerate because their viewport legitimately grows — regenerated on linux to match CI
- Full suite green (verified on linux), `tsc --noEmit` clean

Rotated rect copper pour (was clipped to the un-rotated box):

copper pour rotation


seveibar (approved):

MustafaMulla29 (changes requested):
Remove tests and pull the main and update the repro snapshots(from the repro pr)

abdalraof-albarbar:
done — pulled main (now that #604 merged), removed the tests and just updated the repro snapshots to the fixed renders

tests/pcb/pcb-bounds-plated-hole.test.ts
MustafaMulla29, line 1: Why are you removing the test cases? just make them pass now so that it will cover the cases you fixed here. I meant you to remove the same tests that the repro pr had

abdalraof-albarbar: Okay, I was confused by your words , but I get it now

abdalraof-albarbar:
@MustafaMulla29

MustafaMulla29 (approved):

PR #611, opened 2026-07-08
Include PCB note and Fabrication note text in comprehensive PCB bounds

anil08607:


tests/pcb/pcb-bounds-note-text.test.ts
mohan-bee, line 1: is there any way you can use a real tsx ?

mohan-bee (commented):
what is the issue ?

rushabhcodes (commented):
what are you trying to achieve

lib/pcb/get-text-center-from-anchor-position.ts
imrishabh18, line 1: Doesn't follow the guidelines, export a single method from a file and the file name should match the exported function name

imrishabh18 (changes requested):

imrishabh18 (commented):
I think the changes should be in `core` to expand the board dimension to fit in these texts, this is not the right place.

anil08607:
@imrishabh18 All PCB-bound issues have already been fixed in this repo
.image

PR #613, opened 2026-07-11
add PCB note and Fabrication note text to getComprehensivePcbBounds

anil08607:
This PR updates `getComprehensivePcbBounds` to include both `pcb_note_text` and `pcb_fabrication_note_text` when computing PCB SVG bounds.

Without this, note-only text can render correctly but still get clipped because the viewport calculation ignores its extents. This change makes sure PCB note text contributes to the final bounds, so text annotations stay visible in exported SVGs.


seveibar (approved):

rushabhcodes (commented):
can you use real tsx insted of circuit-json, if we dont provide width and height, it is computed in core ig.

anil08607:
@rushabhcodes  core is also using the Circuit-to-svg for converting the circuit json to svg, if you look into other pcb bound issue. All are handling in this repo.

techmannih (approved):

techmannih:
That's make sense

PR #614, opened 2026-07-11
Add PCB bounds repros for courtyard shapes

anil08607:
## Summary

This PR adds focused visual repro coverage around PCB note text placement and courtyard bounds behavior.

The changes add snapshot-based tests for note text anchor alignment, and introduce one-repro-per-file courtyard cases with inline circuit JSON for `pcb_courtyard_rect`, `pcb_courtyard_circle`, `pcb_courtyard_polygon`, and `pcb_courtyard_outline`. The courtyard repros are tuned so the shapes extend slightly past the board edge, which makes the bounds bug obvious in the rendered snapshot instead of hiding it fully inside the board.

The test inputs were also kept self-contained to make each repro easier to read, tweak, and debug without shared fixture indirection.


seveibar (commented):
i don't think i'm going to approve PRs with this for now, AFAIK this isn't effecting users

PR #616, opened 2026-07-12
Fix: schematic SVG bounds for symbol reference/value text

anil08607:
image

## Summary

- fix schematic SVG bounds so symbol-based `{REF}` and `{VAL}` text contributes to the exported viewport
- prevent long component display names and symbol values from being clipped outside the SVG
- update affected schematic snapshots to match the corrected viewport sizing

## Root Cause

`getSchematicBoundsFromCircuitJson` only accounted for the schematic component body size, while symbol text like reference/designator and value text was rendered separately and could extend past those bounds.

## What Changed

- compute additional bounds for symbol text primitives on schematic components
- resolve `{REF}` from `display_name` / `name`
- resolve `{VAL}` from `symbol_display_value`
- map symbol text positions into real schematic coordinates before expanding bounds


imrishabh18 (commented):
What did this change do? add the before and after snapshot of the fix

anil08607:
@imrishabh18 Done

seveibar (changes requested):
too many snapshots changed (unclear why) - start with a repro

seveibar:
do the repro first

anil08607:
@seveibar I added the repro which has merged by Mustafa
https://github.com/tscircuit/circuit-to-svg/pull/615

lib/sch/get-schematic-bounds-from-circuit-json.ts
```diff
@@ -28,6 +33,7 @@ export function getSchematicBoundsFromCircuitJson(
   let minY = Number.POSITIVE_INFINITY
   let maxX = Number.NEGATIVE_INFINITY
   let maxY = Number.NEGATIVE_INFINITY
+  const circuitSoup = su(soup as any)
```
seveibar: do not use "soup", which is the legacy term for circuit json

lib/sch/get-schematic-bounds-from-circuit-json.ts
```diff
@@ -204,6 +211,122 @@ export function getSchematicBoundsFromCircuitJson(
 
   return { minX, minY, maxX, maxY }
 
+  function updateSymbolTextBounds(component: SchematicComponent) {
```
seveibar: don't use inline closures, this is in our handbook. Review the handbook and add a line to the AGENTS.md

seveibar: https://links.tscircuit.com/

seveibar (commented):
see comments. Do not add so many lines to that entrypoint file

seveibar (approved):

PR #630, opened 2026-07-22
repro: empty <text> element for unresolved schematic symbol placeholder

itisrohit:
Empty SVG `<text>` element is emitted when a schematic symbol placeholder resolves to no value.


seveibar (commented):
? can't understand what is happening with this

itisrohit:
> ? can't understand what is happening with this

This repro is for the case where a schematic symbol placeholder (specifically {VAL}) resolves to no value and the renderer emits an empty SVG <text> element instead of omitting it.

techmannih:
Can you also add fix for that repo in this pr, then we can understand what's should be the excepted solution

itisrohit:
@techmannih I've updated the PR to include the fix as well. The issue is that the renderer was creating a text label even when there was no text to display. Instead of omitting that label, it emitted an empty text element. The fix skips creating text labels when the resolved content is empty, so only labels with actual content are rendered. This keeps the output consistent with the other text renderers in the codebase. Please let me know if this matches the expected behavior or if you'd prefer a different approach.

techmannih:
@itisrohit I am not seeing any changes in the snapshot?

itisrohit:
@techmannih The raw SVG snapshot did change. The fix removes the empty `<text class="sch-component-text"...></text>` element from the generated SVG. Since that element has no content, it doesn't render any pixels, so the visual snapshot remains identical. You should be able to see the change in the `.snap.svg` diff in the PR, where the empty `<text>` element between the `R1` label and the `<circle>` element has been removed.

seveibar:
@itisrohit we can't spend this much time on a bug like this that doesn't effect a user. This time (for both the reviewers and author) should be focused on higher-impact issues

PR #633, opened 2026-07-23
repro: pcb_courtyard_rect missing from getComprehensivePcbBounds

itisrohit:
pcb_courtyard_rect is rendered when `showCourtyards` is enabled, but `getComprehensivePcbBounds()` does not include it. As a result, the SVG viewport transform is computed from the remaining elements only, causing the courtyard to be rendered mostly outside the visible SVG viewport.


techmannih (changes requested):
AFAIK this isn't effecting users, you have try to build board and find issues.

itisrohit:
Thanks for the feedback! I'll close this PR and focus on user-facing issues that can be reproduced by building actual boards.

PR #635, opened 2026-07-25
fix: keep non-finite hole radii out of the SVG (#634)

### LINKED ISSUE #634: Non-finite via/plated-hole diameters render as r="NaN", so the drill silently disappears
### Summary

```tsx
</board>
```

Output:

```html
```

Same for plated holes:

```tsx
```

```html
```

### Cause

```ts
// ...
r: innerRadius.toString(),
```

### Expected

### Related

PR follows.

DPS0340:
Fixes #634.

### What was wrong

A via or plated hole with a non-finite diameter emitted `r="NaN"`:

```html
```

`NaN` isn't a valid SVG length, so renderers discard the `<circle>` entirely — the drill **vanishes silently**. The rest of the board still draws, which makes it worse: you get a plausible-looking PCB with a via that has no hole, and nothing says why.

Both sites multiplied the diameter into a radius and stringified it, and `NaN * n` stays `NaN`.

### A repro note that cost me time — please read before testing

**The value is `NaN` in memory but becomes `null` after JSON serialisation**, and `null * n` is `0`, which renders harmlessly.

So a test built from a hand-written JSON literal with `hole_diameter: null` **passes even without this fix** — I hit exactly that and my first version of these tests was worthless. The bug only appears on the in-memory path `tscircuit` produces, which is the path the converter is normally called on. The tests here use `Number.NaN` for that reason, and I verified they fail without the fix.

### The fix

Clamp at the source in both files:

```ts
const outerRadius = Number.isFinite(rawOuterRadius) ? rawOuterRadius : 0
const innerRadius = Number.isFinite(rawInnerRadius) ? rawInnerRadius : 0
```

Clamping the radius variables themselves (rather than each `.toString()` call site) matters for plated holes, where `maskRadius = outerRadius + soldermaskMargin` feeds two more soldermask circles — those are covered without touching them.

`Number.isFinite` also catches `Infinity`, and follows the pattern already in `lib/sim/simulation-graph-svg/format-value-with-unit.ts` (`if (!Number.isFinite(value)) return "0"`), so this isn't a new convention.

### Scope

Only the two circular-radius paths with a demonstrated failure. I didn't sweep every numeric attribute in the PCB renderer — that's a much larger change, and I'd rather land the proven one first. Happy to follow up if you want the broader pass.

### Tests

`tests/pcb/non-finite-hole-radius.test.ts`, four tests pinning both directions:

| case | expected |
|---|---|
| via, `hole_diameter: NaN` | no `r="NaN"`; outer radius still **> 0**, inner falls back to `0` |
| plated hole, `hole_diameter: NaN` | no `r="NaN"` |
| via, valid `0.6` / `0.3` | both radii > 0 **and their ratio is 2**, matching the diameters |
| plated hole, valid `1` / `0.5` | all radii finite and > 0 |

The ratio assertion is the guard rail: a fix that clamped everything to `0` would pass a bare "no NaN" check but fails this. The first case checks the same thing at finer grain — the valid outer diameter survives while only the unusable drill falls back.

Bite-proofed: reverting `lib/pcb/` takes the file from **4 pass** to **2 pass / 2 fail**.

### Verification

`bun test`: **295 pass / 0 fail** (baseline on `main` is 291 pass / 0 fail; +4 new). No snapshot churn. `bunx tsc --noEmit` clean, `bun run format:check` clean.

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-plated-hole.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-plated-hole.ts
@@ -274,8 +274,15 @@ export function createSvgObjectsFromPcbPlatedHole(
     const scaledHoleWidth = hole.hole_diameter * Math.abs(transform.a)
     const scaledHoleHeight = hole.hole_diameter * Math.abs(transform.a)
 
-    const outerRadius = Math.min(scaledOuterWidth, scaledOuterHeight) / 2
-    const innerRadius = Math.min(scaledHoleWidth, scaledHoleHeight) / 2
+    const rawOuterRadius = Math.min(scaledOuterWidth, scaledOuterHeight) / 2
+    const rawInnerRadius = Math.min(scaledHoleWidth, scaledHoleHeight) / 2
+
+    // A missing or unparseable diameter reaches here as null/NaN, which would be
+    // written as r="NaN" — not a valid SVG length, so renderers drop the circle
+    // and the drill silently vanishes. Clamp at the source so everything derived
+    // from these radii (soldermask mask radius included) stays a valid length.
+    const outerRadius = Number.isFinite(rawOuterRadius) ? rawOuterRadius : 0
+    const innerRadius = Number.isFinite(rawInnerRadius) ? rawInnerRadius : 0
 
     let children: SvgObject[] = [
       {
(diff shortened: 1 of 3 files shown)


DPS0340:
Withdrawing this one on queue grounds rather than anything about the change itself.

I counted my own output across every repo I contribute to. Among PRs a maintainer actually decided, **57% merged** — but **94% of what I have opened has never been looked at by a human at all.** The constraint was never quality; I was handing out more review work than anyone could absorb.

This PR fixes an issue I filed myself, which is the shape most of that volume took: find something, report it, fix it, and hand a maintainer work nobody asked for. I am closing the self-originated ones across every repo and keeping only those that answer someone else's bug report.

The branch stays up and the issue stays open, so nothing is lost. If this one is actually wanted, say so and I will reopen it.

Sorry for the noise.

PR #637, opened 2026-07-25
fix: keep non-finite font sizes and keepout radii out of the SVG (#636)

### LINKED ISSUE #636: Non-finite silkscreen font_size and keepout radius emit font-size="NaN" and r="NaN"
### Summary

Two more places in the PCB renderer write a non-finite number straight into an SVG attribute, producing markup renderers can't use.

**1. Silkscreen text — `font-size="NaN"`**

```tsx
<silkscreentext name="S1" text="hi" fontSize="abc" pcbX={3} pcbY={3} />
```

```html
<text x="0" y="0" fill="#f2eda1" font-family="Arial, sans-serif" font-size="NaN" ...>
```

**2. Circular keepout — `r="NaN"`**

```tsx
<keepout shape="circle" radius="abc" pcbX={5} pcbY={5} />
```

```html
<circle class="pcb-keepout-circle" ... cx="536.36" cy="163.63" r="NaN" .../>
```

Neither is a valid SVG length. The text falls back to the renderer's own default size or vanishes, and the keepout circle is discarded — in both cases the drawing looks fine while quietly showing something other than what was specified.

### Cause

**Silkscreen text** destructures with a default:

```ts
const { font_size = 1, ... } = pcbSilkscreenText
// ...
const transformedFontSize = font_size * scaleFactor
```

A default parameter only covers `undefined`. An unparseable size arrives as `NaN`, passes the default untouched, and `NaN * scaleFactor` stays `NaN`. This is the same shape of bug as `??` not catching `NaN`.

**Keepout** has no guard at all:

```ts
const scaledRadius = circleKeepout.radius * Math.abs(transform.a)
// ...
r: scaledRadius.toString(),
```

`radius` is `null` in the circuit JSON for an unparseable value and `NaN` on the in-memory path.

### Note for reproducing

The value is **`NaN` in memory but `null` after JSON serialisation**, and `null * n === 0`, which renders harmlessly. A fixture built from a hand-written JSON literal with `null` will therefore **not** reproduce either bug — it only shows on the in-memory path `tscircuit` produces, which is how the converter is normally called.

### Expected

Both should degrade to a valid length: font size to the same value the existing `font_size = 1` default produces, and the keepout radius to `0`. Valid values must keep rendering verbatim.

This is the same class as #634 (via / plated hole `r="NaN"`), found by sweeping the remaining numeric attributes in the PCB renderer.

PR follows.

DPS0340:
Fixes #636. Same class as #634, found by sweeping the remaining numeric attributes in the PCB renderer. **Independent of #635** — different files, no conflict, either order.

### What was wrong

Two more non-finite values reaching SVG attributes:

```html
<text ... font-size="NaN" ...>                        <!-- silkscreen text -->
<circle class="pcb-keepout-circle" ... r="NaN" .../>  <!-- circular keepout -->
```

Neither is a valid SVG length. The text falls back to the renderer's own default size or disappears; the keepout circle is discarded outright. Both leave a drawing that looks fine while showing something other than what was specified.

**Silkscreen text is the interesting one.** It already had what looks like a guard:

```ts
const { font_size = 1, ... } = pcbSilkscreenText
const transformedFontSize = font_size * scaleFactor
```

A default parameter only fires on `undefined`. `NaN` sails past it, and `NaN * scaleFactor` stays `NaN` — the same shape as `??` not catching `NaN`.

**Keepout** had no guard at all.

### Repro note — worth knowing before testing

The value is **`NaN` in memory but `null` after JSON serialisation**, and `null * n === 0`, which renders harmlessly. A fixture written as a JSON literal with `null` **passes without the fix**. These tests use `Number.NaN` for that reason, and I confirmed they fail when the change is reverted.

### The fix

- **Silkscreen:** fall back to `scaleFactor`, which is exactly what the existing `font_size = 1` default produces — so a broken size renders at the same size the code already treats as its default, rather than at an arbitrary number.
- **Keepout:** fall back to `0`, matching how #635 handles an unusable radius.

Both use `Number.isFinite` (also catching `Infinity`), consistent with `lib/sim/simulation-graph-svg/format-value-with-unit.ts`. Clamping `scaledRadius` at the variable covers both `r:` sites in the keepout path.

### Tests

`tests/pcb/non-finite-silkscreen-and-keepout.test.ts`, four tests:

| case | expected |
|---|---|
| `font_size: NaN` | no `font-size="NaN"`; every emitted font-size is finite |
| `font_size` 1 vs 2 | rendered size **doubles** — proves scaling is untouched |
| keepout `radius: NaN` | no `r="NaN"` |
| keepout radius 1 vs 2 | rendered radius **doubles**, and is > 0 |

The two ratio tests are the guard rails: a fix that clamped everything to a constant would pass a bare "no NaN" check but fails these. I asserted ratios rather than hardcoded pixel values so the tests don't break on unrelated transform changes.

Bite-proofed: reverting `lib/pcb/` takes the file from **4 pass** to **2 pass / 2 fail**.

### Verification

`bun test`: **295 pass / 0 fail** (291 on `main`; +4 new). No snapshot churn. `bunx tsc --noEmit` clean, `bun run format:check` clean.

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-keepout.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-keepout.ts
@@ -186,7 +186,13 @@ export function createSvgObjectsFromPcbKeepout(
         circleKeepout.center.x,
         circleKeepout.center.y,
       ])
-      const scaledRadius = circleKeepout.radius * Math.abs(transform.a)
+      const rawScaledRadius = circleKeepout.radius * Math.abs(transform.a)
+      // A missing or unparseable radius reaches here as null/NaN and would be
+      // written as r="NaN", which is not a valid SVG length — renderers discard
+      // the circle, so the keepout silently disappears from the drawing.
+      const scaledRadius = Number.isFinite(rawScaledRadius)
+        ? rawScaledRadius
+        : 0
 
       const backgroundAttributes = {
         ...createKeepoutBaseAttributes(
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-silkscreen-text.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-silkscreen-text.ts
@@ -229,7 +229,14 @@ export function createSvgObjectsFromPcbSilkscreenText(
     ]
   }
 
-  const transformedFontSize = font_size * scaleFactor
+  // A default parameter only covers `undefined`; an unparseable size arrives as
+  // NaN and would be written as font-size="NaN", which is not a valid SVG length
+  // — renderers fall back to their own default or drop the text, so the label
+  // silently changes size or disappears.
+  const rawTransformedFontSize = font_size * scaleFactor
+  const transformedFontSize = Number.isFinite(rawTransformedFontSize)
+    ? rawTransformedFontSize
+    : scaleFactor
 
   let textAnchor = "middle"
   let dominantBaseline = "central"
--- a/tests/pcb/non-finite-silkscreen-and-keepout.test.ts
+++ b/tests/pcb/non-finite-silkscreen-and-keepout.test.ts
@@ -1,0 +1,110 @@
+import { expect, test } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "lib"
+
+const board = {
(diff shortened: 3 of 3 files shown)


DPS0340:

Sorry for the volume.

PR #639, opened 2026-07-25
fix: keep non-finite schematic component sizes out of the SVG (#638)

### LINKED ISSUE #638: Non-finite schematic component size renders every coordinate as NaN, so the component vanishes
### Summary

A schematic component whose `size.width` or `size.height` is non-finite renders every coordinate as `NaN`, so the component disappears from the drawing.

```tsx
  <chip name="U1" footprint="soic8" schWidth="abc" />
</board>
```

```html
<rect class="component chip sch-component-body" x="NaN" y="NaN" width="NaN" height="NaN" ...>
```

**112 `NaN` occurrences** in the output for that one chip, spread across `x`, `y`, `width`, `height`, `transform`, `x1`, `y1`, `x2`, `y2`, `cx`, `cy` — the body rect, the pin lines, and the labels. None of those is a valid SVG length, so renderers discard the elements and the chip is simply absent. Other components on the same sheet still draw, which makes it look like a layout problem rather than a bad prop.

### Cause

`lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-box.ts` derives every screen coordinate from the two size numbers:

```ts
const componentScreenTopLeft = applyToPoint(transform, {
  y: schComponent.center.y + schComponent.size.height / 2,
})
```

An unparseable `schWidth` reaches the renderer as `NaN`, and `NaN / 2` stays `NaN`, so it propagates into the top-left, the bottom-right, the width, the height, and everything positioned relative to them.

### Note for reproducing

The value is **`NaN` in memory but `null` after JSON serialisation**, and `null / 2 === 0`, which renders harmlessly. A fixture built from a hand-written JSON literal with `null` will **not** reproduce this — it only shows on the in-memory path `tscircuit` produces, which is how the converter is normally called.

### Scope note

With a real `tscircuit` circuit, `schematic_port.center` also arrives as `{x: null, y: null}` for the affected component — that's upstream data, and a renderer can't invent port coordinates. Guarding the component size fixes the body, the overlay and everything derived from them; isolating that (ports given real coordinates, size left non-finite) takes the output from **8 `NaN`s to 0**.

The upstream half is separate. `tscircuit/core` doesn't error on these values today — I filed the negative-dimension case as tscircuit/core#2861 and the unparseable-unit case as tscircuit/core#2855.

### Expected

A non-finite dimension should degrade to a valid length so the geometry stays well-formed, the same way #634 and #636 handle non-finite radii and font sizes on the PCB side. Valid sizes must keep rendering verbatim.

PR follows.

DPS0340:
Fixes #638. Third in the same sweep as #635 and #637 — **independent of both** (different file, no conflict, any order).

### What was wrong

A schematic component with a non-finite `size` rendered **112 `NaN`s** for a single chip:

```html
<rect class="component chip sch-component-body" x="NaN" y="NaN" width="NaN" height="NaN" ...>
```

They spread across `x`, `y`, `width`, `height`, `transform`, `x1`, `y1`, `x2`, `y2`, `cx`, `cy` — body rect, pin lines and labels. None is a valid SVG length, so renderers discard those elements and the component is simply **absent** from the sheet. Everything else still draws, so it reads as a layout problem rather than a bad prop.

Every screen coordinate is derived from the two size numbers, and `NaN / 2` stays `NaN`:

```ts
```

### Repro note — worth knowing before testing

The value is **`NaN` in memory but `null` after JSON serialisation**, and `null / 2 === 0`, which renders harmlessly. A fixture written as a JSON literal with `null` **passes without the fix**. The tests here use `Number.NaN` for that reason, and I confirmed they fail when the change is reverted.

### The fix

Clamp both dimensions once, at the top, before any coordinate is derived:

```ts
const componentWidth = Number.isFinite(schComponent.size.width) ? schComponent.size.width : 0
const componentHeight = Number.isFinite(schComponent.size.height) ? schComponent.size.height : 0
```

Doing it at the source rather than per-attribute is what makes the body, the overlay and the derived positions all come out well-formed from one guard. `Number.isFinite` also catches `Infinity`, matching #635/#637 and the existing pattern in `lib/sim/simulation-graph-svg/format-value-with-unit.ts`.

### Scope — stated honestly

This does not get a real `tscircuit` circuit all the way to zero `NaN`s, and I don't want to imply otherwise. With `schWidth="abc"`, core also emits `schematic_port.center` as `{x: null, y: null}` for that component; a renderer can't invent port coordinates, so those remain.

What this fixes is measurable in isolation: give the ports real coordinates and leave the size non-finite, and the output goes from **8 `NaN`s to 0**. That's the case the tests pin.

The upstream half is separate and already filed — tscircuit/core#2861 (negative schematic dimensions) and tscircuit/core#2855 (unparseable unit values raising nothing).

### Tests

`tests/sch/non-finite-component-size.test.ts`, four tests:

| case | expected |
|---|---|
| `size.width: NaN` | no `NaN` anywhere in the output |
| `size.height: NaN` | no `NaN` anywhere in the output |
| width 1 vs 2 | rendered body width **doubles** |
| valid 2x1 component | no `NaN` |

The ratio test is the guard rail — a fix that clamped every component to zero would pass a bare "no NaN" check but fails this. I asserted a ratio rather than a pixel constant so it doesn't break on unrelated transform changes.

Bite-proofed: reverting `lib/sch/` takes the file from **4 pass** to **2 pass / 2 fail**.

### Verification

`bun test`: **295 pass / 0 fail** (291 on `main`; +4 new). No snapshot churn. `bunx tsc --noEmit` clean, `bun run format:check` clean.

### DIFF
--- a/lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-box.ts
+++ b/lib/sch/svg-object-fns/create-svg-objects-from-sch-component-with-box.ts
@@ -29,13 +29,25 @@ export const createSvgObjectsFromSchematicComponentWithBox = ({
 }): SvgObject[] => {
   const svgObjects: SvgObject[] = []
 
+  // A missing or unparseable schematic dimension arrives as null/NaN. Every
+  // coordinate below is derived from these two numbers, so letting one through
+  // turns the body rect, the pin lines and the labels into x="NaN" — none of
+  // which is a valid SVG length, so the whole component vanishes from the
+  // drawing. Fall back to 0 so the geometry stays well-formed.
+  const componentWidth = Number.isFinite(schComponent.size.width)
+    ? schComponent.size.width
+    : 0
+  const componentHeight = Number.isFinite(schComponent.size.height)
+    ? schComponent.size.height
+    : 0
+
   const componentScreenTopLeft = applyToPoint(transform, {
-    x: schComponent.center.x - schComponent.size.width / 2,
-    y: schComponent.center.y + schComponent.size.height / 2,
+    x: schComponent.center.x - componentWidth / 2,
+    y: schComponent.center.y + componentHeight / 2,
   })
   const componentScreenBottomRight = applyToPoint(transform, {
-    x: schComponent.center.x + schComponent.size.width / 2,
-    y: schComponent.center.y - schComponent.size.height / 2,
+    x: schComponent.center.x + componentWidth / 2,
+    y: schComponent.center.y - componentHeight / 2,
   })
   const componentScreenWidth =
     componentScreenBottomRight.x - componentScreenTopLeft.x
--- a/tests/sch/non-finite-component-size.test.ts
+++ b/tests/sch/non-finite-component-size.test.ts
@@ -1,0 +1,79 @@
+import { expect, test } from "bun:test"
+import { convertCircuitJsonToSchematicSvg } from "lib"
+
+// NOTE: these use Number.NaN rather than null on purpose. An unparseable
+// dimension reaches the renderer as NaN in memory; it only becomes null once the
+// circuit JSON is serialised, and `null / 2` is 0, which renders harmlessly.
+// A null-based fixture would pass even without the fix.
+
+const circuitWithSize = (size: { width: number; height: number }): any => [
+  {
+    type: "schematic_component",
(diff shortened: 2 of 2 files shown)


DPS0340:

Sorry for the volume.

PR #641, opened 2026-07-25
fix: keep negative lengths out of the SVG (#640)

### LINKED ISSUE #640: Negative dimensions emit negative SVG lengths, so the shape is not rendered
### Summary

Negative dimensions are written straight into SVG length attributes. Per the SVG spec a negative `r`, `width` or `height` is **an error and the shape must not be rendered**, so the element silently disappears rather than looking wrong.

```tsx
  <via name="V1" holeDiameter={-1} outerDiameter={-2} pcbX={3} pcbY={3} />
</board>
```

```html
<circle class="pcb-hole-outer" ... r="-27.272727272727273" .../>
<circle class="pcb-hole-inner" ... r="-13.636363636363637" .../>
```

Four paths are affected:

| input | emitted |
|---|---|
| `<via holeDiameter={-1} outerDiameter={-2} />` | `r="-27.27"`, `r="-13.63"` |
| `<platedhole holeDiameter={-1} outerDiameter={-2} shape="circle" />` | `r="-27.27"` |
| `<hole diameter={-2} />` | `r="-27.27"` |
| `<keepout shape="circle" radius={-1} />` | `r="-27.27"` |

And on the schematic side:

| input | emitted |
|---|---|
| `<chip schWidth={-4} />` | `width="-926.64"` on the component body rect |

The schematic case is slightly different in mechanism: a negative `size.width` flips the two corners, so the derived width comes out negative — same result, the body rect isn't drawn.

The rest of the board renders normally in every case, which makes it look like a placement problem rather than a bad value.

### Cause

Each site multiplies the dimension by the transform scale and stringifies it without checking the sign:

```ts
// create-svg-objects-from-pcb-via.ts
const outerRadius = Math.min(scaledOuterWidth, scaledOuterHeight) / 2
// ...
r: outerRadius.toString(),
```

```ts
// create-svg-objects-from-sch-component-with-box.ts
```

Upstream, `tscircuit/core` accepts these values with no error today — I filed the PCB-side case as tscircuit/core#2857 and the schematic-side case as tscircuit/core#2861. But Circuit JSON can come from several producers, so the renderer shouldn't turn a negative number into markup a renderer refuses to draw.

### Expected

A negative dimension should degrade to a valid length so the geometry stays well-formed, the same way #634/#636/#638 handle non-finite ones. Valid dimensions must keep rendering verbatim.

PR follows.

DPS0340:
Fixes #640.

> Same family as #635 / #637 / #639 (non-finite lengths), but a **different failure mode**: those emitted `NaN`, these emit a well-formed *negative* number that the SVG spec still refuses to draw. This branch is cut from `main` and **does not depend on those PRs** — it touches overlapping files, so if you merge any of them first I'll rebase immediately.

### What was wrong

Per the SVG spec a negative `r`, `width` or `height` is an error and the shape **must not be rendered**. Five paths emitted them:

| input | emitted |
|---|---|
| `<via holeDiameter={-1} outerDiameter={-2} />` | `r="-27.27"`, `r="-13.63"` |
| `<platedhole holeDiameter={-1} shape="circle" />` | `r="-27.27"` |
| `<hole diameter={-2} />` | `r="-27.27"` |
| `<keepout shape="circle" radius={-1} />` | `r="-27.27"` |
| `<chip schWidth={-4} />` | `width="-926.64"` on the body rect |

The rest of the board draws normally, so it reads as a placement problem rather than a bad value.

The schematic case has a different mechanism worth noting: a negative `size.width` **flips the two corners**, so the derived width comes out negative. Clamping the size at the source (rather than the derived width) is what keeps every coordinate below it consistent.

### The fix

One shared helper, `lib/utils/to-svg-length.ts`:

```ts
export const toSvgLength = (value: number): number =>
  Number.isFinite(value) && value > 0 ? value : 0
```

Applied at the five sites where a length is computed — not at each `.toString()` call. That matters in two places: `maskRadius` in both hole renderers is derived from `radius`, so clamping the radius covers the soldermask circles without touching them.

The `Number.isFinite` half also makes this robust to `NaN`/`Infinity`, consistent with the existing pattern in `lib/sim/simulation-graph-svg/format-value-with-unit.ts`.

### Scope

The five paths with a demonstrated failure. I swept the PCB and schematic renderers for negative-length leaks and these were the ones that reproduce end-to-end from a real `tscircuit` circuit; I didn't blanket-wrap every numeric attribute, which would be speculative.

Upstream, `tscircuit/core` accepts these values without error today (filed as tscircuit/core#2857 for the PCB side and tscircuit/core#2861 for the schematic side). Circuit JSON has several producers, so the renderer shouldn't depend on that being fixed first.

### Tests

`tests/pcb/negative-svg-lengths.test.ts`, six tests:

| case | expected |
|---|---|
| via `-2` / `-1` | no negative `r`/`width`/`height` anywhere |
| plated hole `-2` / `-1` | same |
| unplated hole `-2` | same |
| keepout radius `-1` | same |
| schematic component `width: -4` | same |
| valid via `0.6` / `0.3` | both radii **> 0**, and their ratio is **2**, matching the diameters |

The last row is the guard rail: a fix that clamped everything to `0` would pass a bare "no negative lengths" check but fails this.

Bite-proofed: reverting `lib/` takes the file from **6 pass** to **1 pass / 5 fail** — only the valid-input test survives, which is exactly right.

### Verification

`bun test`: **297 pass / 0 fail** (baseline on `main` is 291 pass / 0 fail; +6 new). No snapshot churn. `bunx tsc --noEmit` clean, `bun run format:check` clean.

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-hole.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-hole.ts
@@ -2,6 +2,7 @@ import type { PCBHole } from "circuit-json"
 import { applyToPoint } from "transformation-matrix"
 import type { SvgObject } from "lib/svg-object"
 import type { PcbContext } from "../convert-circuit-json-to-pcb-svg"
+import { toSvgLength } from "lib/utils/to-svg-length"
 
 export function createSvgObjectsFromPcbHole(
   hole: PCBHole,
@@ -26,7 +27,7 @@ export function createSvgObjectsFromPcbHole(
 
   if (hole.hole_shape === "circle" || hole.hole_shape === "square") {
     const scaledDiameter = hole.hole_diameter * Math.abs(transform.a)
-    const radius = scaledDiameter / 2
+    const radius = toSvgLength(scaledDiameter / 2)
 
     if (hole.hole_shape === "circle") {
       const holeElement: SvgObject = {
@@ -375,7 +376,7 @@ export function createSvgObjectsFromPcbHole(
     // Pill shape: two semicircles connected by straight lines
     // If width > height, it's a horizontal pill; if height > width, it's vertical
     const isHorizontal = scaledWidth > scaledHeight
-    const radius = Math.min(scaledWidth, scaledHeight) / 2
+    const radius = toSvgLength(Math.min(scaledWidth, scaledHeight) / 2)
     const straightLength = Math.abs(
       isHorizontal ? scaledWidth - scaledHeight : scaledHeight - scaledWidth,
     )
@@ -494,7 +495,7 @@ export function createSvgObjectsFromPcbHole(
 
     // Same logic as regular pill: handle horizontal and vertical orientations
     const isHorizontal = scaledWidth > scaledHeight
-    const radius = Math.min(scaledWidth, scaledHeight) / 2
+    const radius = toSvgLength(Math.min(scaledWidth, scaledHeight) / 2)
     const straightLength = Math.abs(
       isHorizontal ? scaledWidth - scaledHeight : scaledHeight - scaledWidth,
     )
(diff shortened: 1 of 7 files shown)


DPS0340:

Sorry for the volume.

PR #642, opened 2026-07-27
fix: keep negative and non-finite lengths out of SVG attributes (#640)

LEKKALAGANESH:
Per the SVG spec a negative `r`, `width` or `height` is an error and the shape must not be rendered, so elements silently disappeared instead of merely looking wrong. NaN/Infinity are not lengths at all and had the same effect.

Adds `toSvgLength()` and clamps at the point each length is computed, so every attribute derived from it is covered:

- pcb_via     outer/inner radius
- pcb_hole    circle radius
- pcb_plated_hole  outer/inner radius
- pcb_keepout circle radius
- schematic component body width/height (a negative size.width flips the
  two corners, so the derived screen width came out negative)

Valid dimensions are untouched.

Note: a non-finite dimension also poisons the viewport transform, which leaves NaN in `cx`/`cy` and the board outline. That is a separate defect in the bounds calculation and is not addressed here.

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-hole.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-hole.ts
@@ -1,6 +1,7 @@
 import type { PCBHole } from "circuit-json"
 import { applyToPoint } from "transformation-matrix"
 import type { SvgObject } from "lib/svg-object"
+import { toSvgLength } from "lib/utils/to-svg-length"
 import type { PcbContext } from "../convert-circuit-json-to-pcb-svg"
 
 export function createSvgObjectsFromPcbHole(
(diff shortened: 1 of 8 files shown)


imrishabh18 (commented):
Motivation?

LEKKALAGANESH:
Hi Rishabh,

This PR fixes #640. The core issue is that invalid values reach the SVG
attributes verbatim:

  - `<via holeDiameter={-1} outerDiameter={-2} />→r="-27.27" r="-13.63"`
  - <chip schWidth={-4} /> → width="-926.64"
  - hole_diameter: NaN → r="NaN"

Per the SVG spec, a negative radius, width, or height is an error, causing
the shape to not be rendered. This means the element silently disappears,
making bad input indistinguishable from a missing element during debugging.

While the visible result in a browser remains unchanged (invisible either
way), this fix ensures the emitted document is well-formed. This is
important because these SVGs are also rasterized (e.g., circuit-to-png,
registry image endpoints) and opened in vector editors, where NaN and
negative lengths cause undefined behavior rather than simply skipping the
element.

The change consists of 10 source lines, introducing a toSvgLength() helper
applied where lengths are computed. Tests assert that negative and NaN
values clamp to 0, while valid dimensions remain unchanged.

Please let me know if you would prefer a different approach, such as
emitting a warning instead of clamping, or dropping the element entirely.
If you believe this validation belongs upstream in core rather than the
renderer, I can close this PR.

Best regards,
Lekkala Ganesh

PR #647, opened 2026-07-31
fix: keep non-finite via/plated-hole radii out of the SVG (#634)

### LINKED ISSUE #634: Non-finite via/plated-hole diameters render as r="NaN", so the drill silently disappears

kaivalyaarisetty:
Fixes #634.

### Problem
A via or circular plated hole whose diameter is non-finite renders as `r="NaN"`. A `NaN` SVG length is invalid, so renderers discard the whole `<circle>` — the drill silently disappears while the rest of the board renders fine, so it reads as a placement bug rather than a bad value.

The value is `NaN` in memory but `null` after JSON serialisation (`null / 2 === 0`, which renders harmlessly), so this only shows on the in-memory path `tscircuit` produces — which is how the converter is normally called.

### Fix
- `create-svg-objects-from-pcb-via.ts`: build each circle only when `Number.isFinite(radius)`, so an invalid drill is omitted instead of emitting `r="NaN"`; a finite outer copper ring still renders when only the hole diameter is bad.
- `create-svg-objects-from-pcb-plated-hole.ts` (`shape === "circle"` branch): same finite guard for the outer/inner circles, and the soldermask geometry (derived from `outerRadius`) is skipped when `outerRadius` is non-finite.

Scope is intentionally limited to the non-finite case filed in #634; negative dimensions (#640) are left alone.

### Tests
Adds `tests/pcb/non-finite-hole-diameter.test.ts` covering the via and the circular plated hole (asserts no `NaN` in the output and that the finite outer circle still renders), plus a finite-input case proving unchanged behaviour. Verified the test fails on `main` and passes with this change; the full `tests/pcb` suite (170 tests, 2 snapshots) still passes and `tsc --noEmit` is clean.

---
**Disclosure:** This PR was written by an AI agent (Claude, operating autonomously). The diagnosis, code, and tests were produced and verified by the agent (`bun test` green locally). Happy to adjust the approach — e.g. route through `formatNumber` for an `r="0"` fallback instead of omission — if you prefer a different convention.

### DIFF
--- a/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-plated-hole.ts
+++ b/lib/pcb/svg-object-fns/create-svg-objects-from-pcb-plated-hole.ts
@@ -277,42 +277,54 @@ export function createSvgObjectsFromPcbPlatedHole(
     const outerRadius = Math.min(scaledOuterWidth, scaledOuterHeight) / 2
     const innerRadius = Math.min(scaledHoleWidth, scaledHoleHeight) / 2
 
-    let children: SvgObject[] = [
-      {
-        name: "circle",
-        type: "element",
-        attributes: {
-          class: "pcb-hole-outer",
-          fill: colorMap.copper.top,
-          cx: x.toString(),
-          cy: y.toString(),
-          r: outerRadius.toString(),
-          "data-type": "pcb_plated_hole",
-          "data-pcb-layer": layer,
-        },
-        value: "",
-        children: [],
-      },
-      {
-        name: "circle",
-        type: "element",
-        attributes: {
-          class: "pcb-hole-inner",
-          fill: colorMap.drill,
+    // A non-finite diameter (e.g. an unparseable `outer_diameter`/`hole_diameter`
+    // arriving as NaN) would stringify to r="NaN", which is not a valid SVG
+    // length — renderers discard the whole <circle>. Build only the circles
+    // whose radius is finite instead of emitting broken markup. See
+    // tscircuit/circuit-to-svg#634.
+    const outerCircle: SvgObject | null = Number.isFinite(outerRadius)
+      ? {
+          name: "circle",
+          type: "element",
+          attributes: {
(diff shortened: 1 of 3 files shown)


kaivalyaarisetty:
This is still relevant — the fix for #634 (guarding non-finite via/plated-hole radii out of the SVG) remains valid and CI is green. Happy to rebase onto the latest `main` if that helps, or to add/adjust tests if a maintainer has a preferred approach. Flagging for review when someone has a moment. (Disclosure: this PR was authored by an AI operator.)

kaivalyaarisetty:
Still current and mergeable (CI green). The change guards non-finite via / plated-hole radii (#634) so they're skipped rather than emitting `NaN` into the SVG path data.

One open question for a maintainer: would you prefer I add a small regression test asserting that a via with a non-finite radius produces valid (finite) SVG output, and if so should it live alongside the existing SVG snapshot tests or as a focused unit test? Happy to add it either way once I know the preference.

(Disclosure: authored by an AI operator.)

kaivalyaarisetty:
Still current: rebases cleanly, CI green, mergeable. The #634 fix (skip non-finite via / plated-hole radii instead of emitting `NaN` into SVG path data) is unchanged and low-risk. Keeping this open for whenever a maintainer has a moment to review. (Disclosure: authored by an AI operator.)

kaivalyaarisetty:
Still current and mergeable (CI green). The #634 fix — guarding non-finite via / plated-hole radii so they're skipped instead of emitting `NaN` into SVG path data — is unchanged and low-risk. Keeping this open for a maintainer whenever there's a moment; happy to add a focused regression test or rebase on request. (Disclosure: authored by an AI operator.)

PR #648, opened 2026-07-31
feat: add showSolderPaste option to render solder paste in the pcb view

Hero988:
## What

Adds a `showSolderPaste` option to `convertCircuitJsonToPcbSvg` that renders `pcb_solder_paste` elements in the standard PCB view (off by default, so no existing output changes).

## Why

In tscircuit/core#2945 @techmannih pointed out that solder paste visuals belong in the pcb snapshots rather than a separate snapshot type — but the pcb view currently never draws `pcb_solder_paste`, even though `sort-svg-objects-by-pcb-layer.ts` already reserves layer order `45` for it. This closes that gap so core's pcb snapshot tests can show paste apertures directly.

## How

- `showSolderPaste?: boolean` added to `PcbSvgOptions`/`PcbContext` (same pattern as `showSolderMask`/`showPcbGroups`)
- the element dispatch reuses the existing `createSvgObjectsFromSolderPaste` (already handles rect / rotated_rect / circle / pill / rotated_pill), so the paste look is identical to the solder paste mask view
- layering comes from the pre-existing sort order 45 (paste draws above pads)

## Visual

Same board (rect / circle / pill / 30°-rotated pill pads with 0.7-scale paste), default vs `showSolderPaste: true`:

| default (unchanged) | `showSolderPaste: true` |
| :---: | :---: |
|  |  |

## Tests

- `show-solder-paste-in-pcb-view.test.ts` — snapshot with the option on
- `solder-paste-hidden-by-default-in-pcb-view.test.ts` — snapshot proving default output is unchanged


techmannih:
@Hero988 is this option is not already exist?

seveibar (approved):

seveibar:
@techmannih for your review you can look at the code enough to make that determination

techmannih:
@Hero988 This option already exists in this repo. Could you check again? I think the issue you're facing is in `core`.
 I'm sure it isn't blocking you.

Hero988:
You're right.I checked and `showSolderPaste` is already on `main` (added in #649, released in v0.0.396), so this PR is redundant. Closing it.

Thanks for the pointer on `core` too: with the option available upstream I can wire the core-side PCB snapshots (tscircuit/core#2945, tscircuit/core#2930) straight onto `convertCircuitJsonToPcbSvg({ showSolderPaste: true })` rather than waiting on anything here.

PR #650, opened 2026-08-01
Add repro for pinrow3_smd_ra_male courtyard bounds issue from Circuit

anil08607:

PR #652, opened 2026-08-03
Fix rectangular courtyard PCB bounds

anil08607:
fix #650


lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -203,6 +203,10 @@ export function convertCircuitJsonToPcbSvg(
     },
   }
 
+  const circuitJsonForBounds = options?.showCourtyards
+    ? circuitJson
+    : circuitJson.filter((element) => element.type !== "pcb_courtyard_rect")
```
MustafaMulla29, line 208: Why do you need filtering the circuitjson?

anil08607: @MustafaMulla29 Courtyard bounds should only apply when showCourtyards is true. Existing bounds logic has no rendering options, so filtering here keeps it consistent and prevents hidden courtyards from affecting the viewport.

Without this filtering, hidden courtyards expanded the default viewport and caused snapshot regressions.

anil08607:
We are still on it

MustafaMulla29 (approved):
Take approval from seve as well, am not sure about filtering circuitjson in that file

imrishabh18 (approved):

PR #657, opened 2026-08-07
Replace Game Boy snapshots with a real copper-pour transition

seveibar:
## Stack

This is the reproduction-only base for the fix in #654.

#656 merged the focused reproducer plus Game Boy snapshots. The focused reproducer is valid, but the Game Boy artifact has no nonzero route segment whose copper-pour marker changes between its endpoints, so its snapshots cannot demonstrate this boundary-clipping fix. This PR replaces that non-exercising real-circuit coverage.

## Summary

- remove the Game Boy snapshots that are identical with and without the fix
- add full-board and zoomed snapshots from a published real circuit with actual pour-boundary transitions
- assert that the real fixture contains the route metadata needed to exercise the bug
- make no renderer behavior changes

## Real-circuit reproduction

The fixture is a PCB-only extraction of the published [`ShiboSoftwareDev/solar-battery-charger@1.0.2`](https://tscircuit.com/ShiboSoftwareDev/solar-battery-charger) artifact (release `180c3e01-9b88-49c7-a991-393f0fc45e53`). Only non-PCB elements were removed.

It contains 103 routed traces, 64 copper-pour regions, 257 route points marked inside same-net pours, and—critically—12 real nonzero segments whose route-point markers transition between outside and inside a pour. The test asserts those counts so this fixture cannot silently regress into coverage that does not exercise the bug.

The full-board snapshot is included in the test. This crop makes the failure easy to inspect: the diagonal red trace continues from the black clearance region into the maroon ground pour.

Real circuit trace overlapping its ground pour

## Validation

- real-circuit snapshot tests pass
- `bun run format:check`
- `git diff --check`
- CI runs the complete test and typecheck suites


seveibar:
Closing this replacement: the Game Boy fixture is valid. #654 now uses its autorouted source-net IDs and updates both Game Boy snapshots directly against the reproduction merged in #656.

PR #662, opened 2026-08-08
fix(pcb): fall back to finite bounds when a circuit has no PCB elements

### LINKED ISSUE #661: Circuit with no PCB elements renders a `pcb-boundary` rect with `x="NaN"`
### Summary

`convertCircuitJsonToPcbSvg` emits invalid SVG when the circuit has no elements
that contribute PCB bounds. The `<rect class="pcb-boundary">` comes out with
`x`, `y`, `width` and `height` all set to `"NaN"`, which is not a valid SVG
length. This happens for two valid inputs:

1. an empty circuit `[]`
2. a circuit that only has source/schematic elements, for example a
   schematic-only design or circuit JSON captured before PCB layout has run

Both are valid circuit-json (every element parses against `any_circuit_element`),
so the NaN is the renderer's fault, not bad input.

### Reproduction

```ts
import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"

// 1. empty circuit
convertCircuitJsonToPcbSvg([])

// 2. schematic-only circuit (no PCB elements)
convertCircuitJsonToPcbSvg([
  {
    type: "source_component",
    source_component_id: "sc1",
    name: "R1",
    ftype: "simple_resistor",
    resistance: 1000,
  },
  {
    type: "schematic_component",
    schematic_component_id: "sch1",
    source_component_id: "sc1",
    center: { x: 0, y: 0 },
    size: { width: 1, height: 0.4 },
    rotation: 0,
  },
])
```

### Output

```html
<svg xmlns="http://www.w3.org/2000/svg" width="800" height="600">
  <style></style>
  <rect class="boundary" x="0" y="0" fill="#000" width="800" height="600" .../>
  <rect class="pcb-boundary" fill="none" stroke="#fff" stroke-width="0.3"
        x="NaN" y="NaN" width="NaN" height="NaN"
        data-type="pcb_boundary" data-pcb-layer="global"/>
</svg>
```

### Cause

`getComprehensivePcbBounds` starts `minX`/`minY` at `+Infinity` and
`maxX`/`maxY` at `-Infinity` and returns them unchanged with `hasBounds: false`
when nothing updates them. `convertCircuitJsonToPcbSvg` does not look at
`hasBounds`. It computes `circuitWidth = maxX - minX + 2 * padding`, which is
`-Infinity`. The centering offset `(svgWidth - circuitWidth * scaleFactor) / 2`
then works out to `Infinity * 0`, which is `NaN`. That NaN transform is applied
to the boundary corners, so the rect renders as `NaN`.

The schematic converter already handles this: `getSchematicBoundsFromCircuitJson`
falls back to a finite default box (`if (!Number.isFinite(minX)) { minX = -1; maxX = 1 }`)
so an empty schematic renders a valid SVG. The PCB path has no equivalent.

This is the same "NaN in the SVG" family as #634, #636 and #638, but the trigger
is a valid finite circuit with no PCB extent rather than a non-finite field value.

zkasuran:
Fixes #661.

### Problem

`convertCircuitJsonToPcbSvg` emits an invalid `<rect class="pcb-boundary">` with
`x="NaN" y="NaN" width="NaN" height="NaN"` when the circuit has no elements that
contribute PCB bounds. Two valid inputs hit this:

1. an empty circuit `[]`
2. a circuit that only has source/schematic elements (a schematic-only design,
   or circuit JSON captured before PCB layout)

`getComprehensivePcbBounds` returns non-finite bounds (`Infinity`) with
`hasBounds: false` for these. The converter fed those infinities straight into
its scale/translate transform. `circuitWidth * scaleFactor` works out to
`Infinity * 0 = NaN`, so every projected coordinate became `NaN`.

### Fix

Honor the `hasBounds` flag the bounds function already returns. When bounds are
absent or non-finite, fall back to a small finite box before building the
transform. This mirrors `getSchematicBoundsFromCircuitJson`, which already does
`if (!Number.isFinite(minX)) { minX = -1; maxX = 1 }` so an empty schematic
renders a valid SVG. Circuits that do have PCB bounds are unaffected: the guard
only runs when the transform would otherwise be non-finite.

Scope is limited to this no-bounds case. Non-finite and negative field values on
individual elements (#634, #636, #638, #640) are separate and left alone.

### Tests

`tests/pcb/pcb-empty-circuit-nan-bounds.test.ts`:

- asserts the empty and schematic-only inputs are valid circuit-json (each parses
  against `any_circuit_element`) and that `getComprehensivePcbBounds` reports
  `hasBounds: false` for them
- `convertCircuitJsonToPcbSvg([])` contains no `NaN` or `Infinity`
- the schematic-only circuit renders a `pcb-boundary` rect with finite
  `x`/`y`/`width`/`height`, plus an SVG snapshot of the now-valid output

Verified the two render assertions fail on `main` (the output has `x="NaN"`) and
pass with this change. Full `bun test` is 308 pass / 1 fail, where the single
failure is the pre-existing font-dependent `pcb-pad-trace-clearance-error` image
snapshot that also fails on `main` and is unrelated to this change. No other
committed snapshot changed. `bunx tsc --noEmit` clean and `biome format` clean on
the touched files.

---

AI assistance (Claude, Anthropic) was used in developing this change. The design,
review and verification were done by the author. Verified locally before
submitting: `bun test` (308 pass, 1 pre-existing unrelated flaky fail),
`bunx tsc --noEmit` and `biome format` on the changed files.

### DIFF
--- a/lib/pcb/convert-circuit-json-to-pcb-svg.ts
+++ b/lib/pcb/convert-circuit-json-to-pcb-svg.ts
@@ -216,17 +216,31 @@ export function convertCircuitJsonToPcbSvg(
     ? circuitJson
     : circuitJson.filter((element) => element.type !== "pcb_courtyard_rect")
 
-  const {
-    minX,
-    minY,
-    maxX,
-    maxY,
-    boardMinX,
-    boardMinY,
-    boardMaxX,
-    boardMaxY,
-    hasBoardBounds,
-  } = getComprehensivePcbBounds(circuitJsonForBounds)
+  const pcbBounds = getComprehensivePcbBounds(circuitJsonForBounds)
+  const { boardMinX, boardMinY, boardMaxX, boardMaxY, hasBoardBounds } =
+    pcbBounds
+
+  // getComprehensivePcbBounds returns non-finite bounds (Infinity) when nothing
+  // contributes PCB bounds. That happens for an empty circuit or for one that
+  // only has source/schematic elements (a schematic-only design or circuit JSON
+  // captured before PCB layout). Those infinities flow into the scale/translate
+  // transform below and turn every projected coordinate into NaN, so the SVG
+  // ends up with an invalid pcb-boundary rect whose x/y/width/height are "NaN".
+  // Fall back to a small finite box the way getSchematicBoundsFromCircuitJson
+  // does for the empty schematic so the PCB SVG stays valid.
+  let { minX, minY, maxX, maxY } = pcbBounds
+  if (
+    !pcbBounds.hasBounds ||
+    !Number.isFinite(minX) ||
+    !Number.isFinite(minY) ||
+    !Number.isFinite(maxX) ||
+    !Number.isFinite(maxY)
+  ) {
+    minX = -1
+    minY = -1
+    maxX = 1
+    maxY = 1
+  }
 
   const { boundsMinX, boundsMinY, boundsMaxX, boundsMaxY, padding } =
     getViewportBounds({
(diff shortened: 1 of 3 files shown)


imrishabh18 (changes requested):
Not a valid issue, you can't create a circuit without a board or group. Which has bounds calculation already

zkasuran:
Thanks for the review, that is a fair point. In the normal flow a board or group always provides PCB bounds, so the NaN rect only shows up for hand-built circuit JSON that has neither, which is not real usage. I will close this and #661. Appreciate you looking.

PR #676, opened 2026-08-20
render schematic sheet frames at their declared centers

mohan-bee:
motivation: components overlapping or moving off from the frame 

before: 

image

after: 

image


seveibar (changes requested):
!! totally incorrect behavior, circuit json always renders things where they're specified in circuit json

lib/sch/schematic-sheet-utils.ts
```diff
@@ -23,14 +25,18 @@ export interface SchematicSheetLayout {
   innerMaxY: number
 }
 
+export type SchematicSheetWithCenter = SchematicSheet & {
+  center?: Point
```
seveibar: why would you add this if it isn't in the spec!!!

PR #686, opened 2026-08-25
Render schematic pin-label font size overrides

GokulPandi-M:
## Summary

Uses the optional numeric `schematic_port.display_pin_label_font_size` value when rendering schematic pin labels.

The existing renderer defaults remain unchanged when the field is omitted:

- normal pin labels continue to use `0.15mm`
- negated and inversion-circle labels continue to use `0.12mm`
- an explicit numeric value overrides either default

## Why

This completes the renderer layer for per-port pin-label sizing. It lets producers provide a precise font size for crowded imported symbols without scaling the symbol or adding renderer-side collision heuristics.

Circuit JSON support was added in https://github.com/tscircuit/circuit-json/pull/723.
The Core integration is https://github.com/tscircuit/core/pull/3430.

## Scope

The production change only passes the validated Circuit JSON value to the existing schematic font-size helper. It does not add label measurement, collision detection, part-specific values, or new rotation logic.

## Validation

- full test suite: 323 passed, 0 failed
- focused tests cover omitted defaults, numeric overrides, negated labels, inversion-circle labels, and vertical label rotation
- TypeScript check passes
- package build passes
- changed files pass Biome formatting

The repository-wide Biome command also reports the existing oversized `tests/pcb/assets/abse-gameboy-1.0.16-pcb.json` fixture; it reports no formatting changes.


anil08607 (commented):
No snapshot?

ShiboSoftwareDev (approved):

imrishabh18 (approved):

GokulPandi-M:
@imrishabh18 @ShiboSoftwareDev can you merge please becoz i have another pr which depends on this

PR #697, opened 2026-08-26
Render configurable schematic sheet dimensions

Sang-it:
## Summary

- render A4 by default and add ANSI B schematic sheet frames
- let explicit sheet_width and sheet_height override the named sheet size per dimension
- apply dimensions to sheet frames, bounds, and stacked-sheet layout
- update circuit-json to the merged schema release

## Verification

- bun test (326 passed)
- bunx tsc --noEmit
- bun run build
- bun run format:check
- git diff --check


tests/sch/schematic-sheet-explicit-dimensions.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  expect(custom.height * SCHEMATIC_UNIT_TO_MM).toBeCloseTo(300)
+
+  const circuitJson: AnyCircuitElement[] = [
+    {
+      type: "schematic_sheet",
+      schematic_sheet_id: "schematic_sheet_custom",
+      name: "Custom Sheet",
+      sheet_size: "a4",
+      sheet_width: 500,
+      sheet_height: 300,
```
seveibar: this is invalid data, not a good test

tests/sch/schematic-sheet-size.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  expect(a4.height * SCHEMATIC_UNIT_TO_MM).toBeCloseTo(210)
+  expect(ansiB.width * SCHEMATIC_UNIT_TO_MM).toBeCloseTo(431.8)
+  expect(ansiB.height * SCHEMATIC_UNIT_TO_MM).toBeCloseTo(279.4)
+
+  const circuitJson: AnyCircuitElement[] = [
+    {
+      type: "schematic_sheet",
+      schematic_sheet_id: "schematic_sheet_ansi_b",
+      name: "ANSI B Sheet",
+      sheet_size: "ansi_b",
```
seveibar: sheet_width/sheet_height should always be provided, the lookup can be a fallback but sheet_width/sheet_height should be provided

lib/sch/get-schematic-bounds-from-circuit-json.ts
```diff
@@ -34,7 +34,11 @@ export function getSchematicBoundsFromCircuitJson(
   // Find the bounds
   for (const item of circuitJson) {
     if (item.type === "schematic_sheet") {
-      const layout = getSchematicSheetLayout()
+      const layout = getSchematicSheetLayout(
+        item.sheet_size,
+        item.sheet_width,
+        item.sheet_height,
```
seveibar: 2 parameter rule, you could simply pass item here

seveibar (commented):
see comments

tests/sch/schematic-sheet-explicit-dimensions.test.ts
Sang-it: Fixed: this test now uses explicit sheet dimensions without a conflicting named size.

tests/sch/schematic-sheet-size.test.ts
Sang-it: Fixed: the named-size fixtures now include the resolved sheet_width and sheet_height values.

lib/sch/get-schematic-bounds-from-circuit-json.ts
Sang-it: Fixed: getSchematicSheetLayout now takes the schematic sheet object directly, and all renderer call sites pass the item through.

seveibar (approved):

PR #699, opened 2026-08-28
Render polygon solder paste apertures

ShiboSoftwareDev:
## Why

Real Altium boards use non-rectangular stencil apertures. After import, those apertures are represented as polygonal `pcb_solder_paste` elements, but Circuit-to-SVG could not draw that shape in either the PCB view or the dedicated solder-paste mask.

## What changed

- Render polygonal solder-paste apertures from their actual points
- Include polygon points when calculating solder-paste-mask bounds
- Keep layer filtering and the existing solder-paste color behavior
- Add one focused SVG snapshot test in its own file

This is stacked on tscircuit/circuit-json#738, which adds the polygon variant to the existing `pcb_solder_paste` schema.

## Validation

- Focused polygon SVG snapshot test
- Existing rectangle, circle, oval, pill, and rotated-pill solder-paste tests
- `bunx tsc --noEmit`
- `bun run build`
- Biome formatting on the changed files


imrishabh18 (approved):

imrishabh18 (commented):
Check the CI

PR #701, opened 2026-08-29
match pcb view font in snspshot

mohan-bee:
### Motivation
 render pcb snapshot as exactly like pcb view

PCB VIEW: 

image

CURRENT VIEW: 

image

AFTER FIX: 

image


tests/pcb/__snapshots__/silkscreen-knockout-basic.snap.svg
mohan-bee, line 1: knockout pcb view

Image

seveibar (changes requested):
This should be a default-off setting for svgs

it increases the size of svg and decreases readability- ie it is kind of intentional. Tbh im not sure we need it- there is corcuit to canvas for accurate text renders

PR #704, opened 2026-08-31
Fix non-finite via and plated-hole diameters emitting r="NaN"

### LINKED ISSUE #634: Non-finite via/plated-hole diameters render as r="NaN", so the drill silently disappears

marcos452652258-gif:
Fixes #634

### Summary
- Guard against non-finite/NaN `outer_diameter` and `hole_diameter` when computing radius values for PCB vias and plated holes across PCB, Assembly, and Pinout rendering modules.
- Safely parse optional distance values in PCB bounds calculation to prevent Zod errors on non-finite dimensions.
- Added comprehensive tests verifying that non-finite/NaN dimensions do not emit `r="NaN"` attributes.

### DIFF
--- a/lib/assembly/svg-object-fns/create-svg-objects-from-assembly-plated-hole.ts
+++ b/lib/assembly/svg-object-fns/create-svg-objects-from-assembly-plated-hole.ts
@@ -81,10 +81,18 @@ export function createSvgObjectsFromAssemblyPlatedHole(
 
   // Fallback to circular hole if not pill-shaped
   if (hole.shape === "circle") {
-    const scaledOuterWidth = hole.outer_diameter * Math.abs(transform.a)
-    const scaledOuterHeight = hole.outer_diameter * Math.abs(transform.a)
-    const scaledHoleWidth = hole.hole_diameter * Math.abs(transform.a)
-    const scaledHoleHeight = hole.hole_diameter * Math.abs(transform.a)
+    const scaledOuterWidth =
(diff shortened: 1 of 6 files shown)


marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

PR #708, opened 2026-09-04
fix: render overlined pin labels and schematic text

hrithik18k:
KiCad active-low pin names and local wire labels are represented as structured text parts, but the SVG renderer previously understood only the legacy whole-label `N_` convention. This change renders `display_pin_label_text_parts` and `schematic_text.text_parts` as native SVG `<tspan>` elements, applying `text-decoration="overline"` only to parts marked `is_overlined`. A mixed label such as `RI/CLK` therefore receives one continuous overline over `RI`, while `/CLK` remains plain. A shared text-part renderer handles both fields without manual line geometry or positioning constants, while unstyled text and legacy `N_` labels retain their existing behavior.

**Motive:** Preserve KiCad active-low formatting through Circuit JSON and render it faithfully in schematic SVG output.

## Reproduction

The baseline reproduction was merged in [#721](https://github.com/tscircuit/circuit-to-svg/pull/721) using the unmodified Easyduino Circuit JSON generated by `kicad-to-circuit-json`. This PR updates that same test and snapshot to verify the fix: KiCad `~{...}` markup is absent, all 15 structured overline parts are rendered, and the partial `RI/CLK` label overlines only `RI`.

## Validation

- Uses the released `circuit-json@0.0.486` structured text-part schema from [circuit-json #769](https://github.com/tscircuit/circuit-json/pull/769)
- 357 tests pass; 1 existing test remains TODO
- Package and TypeScript declarations build successfully
- Changed files pass Biome formatting and `git diff --check`


imrishabh18 (changes requested):
snapshots?

tests/sch/overlined-pin-label.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+        element.type === "schematic_port" &&
+        element.display_pin_label === "ABCD",
+    )
+
+    if (!activeLowPort) throw new Error("Expected the pin 2 schematic port")
+    activeLowPort.display_pin_label = "RI/CLK"
+    activeLowPort.display_pin_label_text_parts = [
+      { text: "RI", is_overlined: true },
+      { text: "/CLK" },
+    ]
```
imrishabh18: Why would you hack and override the circuit json? this is wrong. Just use the generated circuit-json

imrishabh18 (changes requested):

tests/sch/overlined-pin-label.test.tsx
hrithik18k: Resolved — I replaced the TSX-generated-and-mutated data with a typed Circuit JSON fixture containing display_pin_label_text_parts directly, then regenerated the focused snapshot. The renderer test now consumes the same structure produced by kicad-to-circuit-json.

MustafaMulla29 (commented):
First repro then fix, you should use real example as repro.

techmannih (changes requested):
why did you not added the repro before a fix?

hrithik18k:
Added the requested reproduction before the fix: https://github.com/tscircuit/circuit-to-svg/pull/721. It uses the full generated Easyduino Circuit JSON without mutating it in the test, and the committed SVG snapshot shows the missing active-low overlines. #708 remains the follow-up renderer fix.

MustafaMulla29 (approved):
Get review from others as well, LGTM!

tests/__snapshots__/easyduino-overline-repro.snap.svg
techmannih, line 1: @hrithik18k I am not seeing any difference in the snapshot?

hrithik18k: @techmannih 
before : 
image

after :
image

techmannih: Look what the hack, in the swiping there is no difference but in 2-up difference shown
Image
Image

techmannih: @hrithik18k Yeah, I see the 2-up difference clearly. I didn’t notice it at first; I noticed it later.

techmannih (approved):

imrishabh18 (approved):

PR #721, opened 2026-09-10
repro: active-low schematic text loses overlines

hrithik18k:
The current renderer ignores structured overline parts on `schematic_text`, so active-low pin names and local wire labels from the Easyduino schematic are displayed as plain text. This reproduction renders the complete Circuit JSON generated from `Easyduino_ESP32.kicad_sch` without changing the JSON inside the test.

## Current behavior

The current `main` output has no overlines on the active-low pin names or the external wire labels:

Current circuit-to-svg output without active-low overlines

## Intended result

The proposed renderer output adds continuous overlines to the marked text parts:

Proposed circuit-to-svg output with active-low overlines

In the intended image, inspect both kinds of text around pins 24, 28, and 11:

- the teal pin names inside the component: `RTS`, `DTR`, and `SUSPEND`
- the black external wire labels: `RTS`, `DTR`, and `SUSPEND`

Both sets have overlines. The change also overlines only `RI` in `RI/CLK` and only the active-low prefixes in labels such as `TXT/GPIO.0`; suffixes such as `/CLK` and `/GPIO.0` remain plain. The expected output follows the KiCad rendering in the [Easyduino conversion reference](https://raw.githubusercontent.com/tscircuit/kicad-to-circuit-json/main/tests/repros/easyduino-overline/__snapshots__/easyduino-overline.snap.png).

**Motive:** Capture the existing rendering failure with a real generated schematic before applying the renderer fix.

## Validation

- `bun test tests/easyduino-overline-repro.test.ts`
- `bunx tsc --noEmit`
- the generated JSON fixture passes Biome formatting

Follow-up fix: [#708](https://github.com/tscircuit/circuit-to-svg/pull/708).


techmannih (commented):
please a reference in pr description

techmannih:
add what we have now in snapshot, and clearly point out what you want to add.

hrithik18k:
Updated the PR description with a side-by-side comparison of the current `main` snapshot and the intended overlined output. I also listed exactly which label portions should gain an overline and which suffixes must remain plain.

techmannih:
@hrithik18k I got it. im not sure we have to support this you have to confirm with @seveibar

seveibar (approved):

seveibar:
@techmannih yes this is a bit weird but the circuit json iiuc comes from kicad, so it's not supported in TSX (yet), @hrithik18k fwiw getting support in TSX is the more "normal path" and makes the repros more grounded

PR #741, opened 2026-09-12
test: reproduce fabrication note visibility and layer leaks

anil08607:
Adds four SVG snapshot repros for fabrication annotations remaining visible with `showPcbNotes: false` and top-only fabrication paths appearing on the bottom face.

Covers top/bottom rendering with PCB notes enabled and disabled. Snapshots intentionally capture the current incorrect behavior for a follow-up fix.

Related issue: https://github.com/tscircuit/circuit-json-to-gltf/issues/199

Validation:
- 4 repro tests pass.
- Full suite: 363 pass, 0 fail, 1 existing TODO.


anil08607:
@raykholo, I'm adding the repro now and will fix it in another PR. Could you take a look at it once?

tests/pcb/pcb-fabrication-note-visibility.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    to: { x: 4, y: -4 },
+    text: "3 mm",
+    font: "tscircuit2024",
+    font_size: 1,
+    arrow_size: 0.5,
+  },
+] satisfies CircuitJson
+
+for (const layer of ["top", "bottom"] as const) {
+  for (const showPcbNotes of [false, true]) {
```
imrishabh18: Don't add this pattern, use individual `test` and explain what you are testing

imrishabh18 (changes requested):

seveibar (commented):
can you use TSX for the repro

seveibar (changes requested):
unclear from repros what you're fixing, one test per file

seveibar (changes requested):
what are you testing? It's not clear from the repro, if you're introducing a repo it should be clear looking at the repro what you're testing

CleanShot 2026-09-14 at 07 31 13@2x

PR #752, opened 2026-09-14
Resolve board defaults for via tenting

techmannih:
Standalone and trace-route vias inherit omitted tenting fields from their owning board, preserving explicit per-side and legacy overrides. Board ownership is computed once from element and subcircuit relationships and passed through `PcbContext.boardOwnerMap` for ID lookups during rendering.

One labelled snapshot shows both faces of a two-board panel with opposite defaults, explicit overrides, and standalone/route-via deduplication. Three focused tests cover panel rendering, ownership independent of board positions and element order, and rendering without a board.

Validation: 384 tests pass, 1 existing todo, 0 failures; typecheck, build, and format check pass.

Uses https://github.com/tscircuit/circuit-json/pull/796 for https://github.com/tscircuit/core/pull/3926.


lib/pcb/svg-object-fns/create-svg-objects-from-pcb-via.ts
```diff
@@ -14,10 +14,16 @@ export function createSvgObjectsFromPcbVia(
     ctx.showSolderMask && (layer === "top" || layer === "bottom")
   if (showSolderMask && !hole.layers.includes(layer)) return []
 
+  const board = ctx.circuitJson?.find(
+    (element): element is PcbBoard => element.type === "pcb_board",
```
seveibar: this is a bug, it will break if there are panels

seveibar (commented):
use a single test snapshot, make it clear from the image what the values are etc.

as-is, the visual snapshot is insufficient to evaluate correctness

seveibar (changes requested):
Sloppy board ownership determination, no geometric matching should be needed and it should be pre computed and passed as part of the context eg boardOwnerMap.get(anyId)

tests/pcb/pcb-via-board-tenting.fixture.ts
```diff
@@ -0,0 +1,230 @@
...
+    legacy,
+    note(`Expect ${inheritedState}`, x - 39, -1),
+    note("Expect EXPOSED", x - 13, -1),
+    note("Expect TENTED", x + 13, -1),
+    note("Expect EXPOSED", x + 39, -1),
+    note("pcb_trace.route vias (same nested group)", x, -8, 2.4),
+    note("top: unset", x - 39, -12),
+    note("bottom: unset", x - 39, -15),
+    note("top: false", x - 13, -12),
+    note("bottom: false", x - 13, -15),
```
imrishabh18: What the hell is this pattern??! Have we ever followed something like this 🤦🏻‍♂️

imrishabh18 (changes requested):

lib/pcb/convert-circuit-json-to-pcb-svg.ts
```diff
@@ -153,6 +155,7 @@ export interface PcbContext {
   showAnchorOffsets?: boolean
   showPinNumbers?: boolean
   circuitJson?: AnyCircuitElement[]
+  boardOwnerMap?: Map<string, PcbBoard | undefined>
```
seveibar:
```suggestion
  boardOwnerMap?: Map<AnyCircuitJsonId, PcbBoard | undefined>
```

then in create-board-ownermap define:

```tsx
type AnyCircuitJsonId = string
```

We never use `Map<string,...`

seveibar (approved):
see comment

lib/pcb/create-board-owner-map.ts
```diff
@@ -0,0 +1,64 @@
...
+    const parentId = [
+      "subcircuit_id" in element && element.subcircuit_id,
+      "pcb_group_id" in element && element.pcb_group_id,
+      "pcb_component_id" in element && element.pcb_component_id,
+      "pcb_trace_id" in element && element.pcb_trace_id,
+      "source_group_id" in element && element.source_group_id,
+      "parent_source_group_id" in element && element.parent_source_group_id,
+      "parent_subcircuit_id" in element && element.parent_subcircuit_id,
+    ].find((parentId) => parentId && parentId !== id)
+    parentById.set(id, parentId || undefined)
```
imrishabh18: This section seems like some kind of hack!

imrishabh18 (changes requested):

imrishabh18 (approved):

PR #754, opened 2026-09-15
feat: support ccw_rotation for hole_with_polygon_pad plated holes (#709)

### LINKED ISSUE #709: hole_with_polygon_pad plated holes are drawn unrotated: ccw_rotation is ignored for the pad outline and the drill offset
## Summary

`create-svg-objects-from-pcb-plated-hole.ts` renders a `hole_with_polygon_pad` plated hole by translating `pad_outline` and the hole offset by `(x, y)` only. The `ccw_rotation` field that circuit-json defines for this shape is never read, and the same happens in `get-pcb-bounds-from-circuit-json.ts` and in the pin-number geometry.

core derives the rotation of every plated hole from the component transform (`rect_ccw_rotation`, `hole_ccw_rotation`, `ccw_rotation`); for this shape it is being fixed to emit `ccw_rotation` too (tscircuit/core#3647). Once that lands, a `<chip pcbRotation={90}>` with polygon-pad holes renders its pads unrotated in the SVG, with the drill offset on the wrong side, while its rect-pad holes rotate correctly.

## Repro

A 6mm x 1mm bar (`pad_outline` from (-3,-0.5) to (3,0.5), `hole_offset_x: 2`) with `ccw_rotation: 90` still produces a horizontal `<polygon class="pcb-hole-outer-pad">` and a drill to the right of the pad center. Expected: a vertical bar with the drill above the center.

## Expected

Rotate `pad_outline` and the hole offset around the hole position by `ccw_rotation` (default 0), like the oval/pill and rect-pad branches already do with their rotation fields.

Ialyahya96:
Fixes #709

## Summary

`hole_with_polygon_pad` plated holes previously ignored `ccw_rotation` defined in `circuit-json`, translating `pad_outline` and the hole offset by `(x, y)` without applying rotation. This resulted in unrotated polygons and drill offsets placed at incorrect coordinates when rotation was applied (e.g. via `<chip pcbRotation={...}>`).

This PR adds full `ccw_rotation` support across all relevant modules:
1. **SVG Plated Hole Rendering** (`lib/pcb/svg-object-fns/create-svg-objects-from-pcb-plated-hole.ts`):
   - Rotates `pad_outline` points by `ccw_rotation` counter-clockwise around the local origin `(0, 0)`.
   - Rotates drill offsets `(hole_offset_x, hole_offset_y)` by `ccw_rotation`.
   - Propagates rotation to inner drill shapes (`oval`, `pill`, `rotated_pill`) via SVG rotation transform.
2. **PCB Bounds Calculation** (`lib/pcb/get-pcb-bounds-from-circuit-json.ts`):
   - Computes rotated coordinates for `pad_outline` points before expanding bounding box bounds to avoid SVG viewport clipping.
3. **Pin Number Centroid & Text Rotation** (`lib/pcb/svg-object-fns/create-svg-object-from-pcb-pad-pin-number.ts`):
   - Rotates polygon centroid offset by `ccw_rotation` around hole center.
   - Passes `ccwRotation` to pin number geometry.

## Repro & Verification

Added test suite in `tests/pcb/pcb-plated-hole-polygon-pad-rotation.test.ts`:
- Verified unrotated control vs 90° rotation matches expected geometry:
  - Repro bar `[-3, -0.5]` to `[3, 0.5]` with `hole_offset_x: 2`:
    - 0°: pad width 6mm, height 1mm, drill at `(12, 10)` (right of center `(10, 10)`).
    - 90° CCW: pad width 1mm, height 6mm, drill at `(10, 12)` (above center `(10, 10)` in SVG coordinates).
- Verified non-orthogonal rotations (45°), 180°, 270°.
- Verified drill shape transformations (`pill`, `oval`, `rotated_pill`).
- Verified PCB bounds calculation matches rotated geometry.
- Checked full test suite with `bun test` (all tests pass).
- Formatted with Biome (`bun run format`) and built (`bun run build`).


rushabhcodes:
any real use case for this?

Ialyahya96:
Hi @rushabhcodes! Great question. Yes, there are several concrete, real-world hardware and EDA use cases for rotated polygon plated holes:

1. **Angled High-Current Connectors & Heavy Mechanical Components:**
   Components like angled barrel jacks, heavy relays, automotive terminal blocks, RF shield clips, and USB-C shell tabs often use non-standard polygon or teardrop-reinforced plated through-holes to maximize copper annular rings for mechanical shear resistance and current density. When these components are placed at angles (e.g. 45°, 90°, or along rotated board contours), their plated holes must rotate with the component footprint.

2. **Asymmetric Hole Offsets:**
   Many polygon plated pads intentionally offset the drill from the pad centroid (for example, keyway alignment pins, edge connectors, or teardrop drill entries). When `ccw_rotation` is applied, both the polygon outline and the `hole_offset_x`/`hole_offset_y` must rotate together. Without this fix, the outer copper pad stays unrotated or the drill offset disconnects from the pad orientation, distorting the physical annular ring.

3. **KiCad & CAD Round-Trip Fidelity:**
   Custom footprint converters (like `circuit-json-to-kicad` and KiCad footprint importers) emit custom polygon pads as `hole_with_polygon_pad`. If a rotated board or footprint is rendered in SVG or previewed in web documentation without rotation, the visual SVG diverges from the fabricated Gerber / KiCad output.

4. **Consistency Across the Renderer:**
   Standard pads (`rect`, `oval`, `pill`) already support `ccw_rotation` in `circuit-to-svg`. This PR brings `hole_with_polygon_pad` into full feature parity so rotation behaves predictably across all plated hole geometries.

Happy to make any adjustments or add more snapshot fixtures if you have specific preferences!

PR #755, opened 2026-09-15
fix: include rotated dimension labels in PCB bounds

santosh-kumar-git:
A 90° label on a top-edge PCB dimension is clipped by the automatically calculated SVG viewport. This affects exported PCB notes and fabrication dimensions: the dimension line is visible, but part of its measurement text falls outside the image. I found this while checking the bounds against the existing dimension renderers; I do not have a separate reported user issue.

Reproduce on the PR base, `7c9208dc8ce7e359fd8b407cf0bdd1d4bf196538` (v0.0.423). In the repository, save this as `repro-dimension.ts` and run `bun repro-dimension.ts`, then open `dimension.svg`:

```ts
import type { AnyCircuitElement } from "circuit-json"
import { convertCircuitJsonToPcbSvg } from "./lib"

const circuit: AnyCircuitElement[] = [
  {
    type: "pcb_board", pcb_board_id: "board",
    center: { x: 5, y: -2 }, width: 10, height: 4,
    num_layers: 2, material: "fr4", thickness: 1.6,
  },
  {
    type: "pcb_note_dimension", pcb_note_dimension_id: "dimension",
    layer: "top", from: { x: 0, y: 0 }, to: { x: 10, y: 0 },
    text: "10.00 mm", font: "tscircuit2024", font_size: 1,
    arrow_size: 0.5, offset_distance: 1,
    offset_direction: { x: 0, y: 1 }, text_ccw_rotation: 90,
  },
]
await Bun.write("dimension.svg", convertCircuitJsonToPcbSvg(circuit, {
  width: 400, height: 300,
}))
```

Expected: the complete vertical `10.00 mm` label fits inside the 400×300 image. On the base, the top of the label is cropped. The dimension bounds stop at y=2.25 mm although the renderer's existing clearance and rotation place the label's upper extent at y=6.85 mm. With this PR, the automatic bounds contain the whole label. The same input using `pcb_fabrication_note_dimension` reproduces the fabrication case.

The fix makes bounds account for the clearance and combined segment/text rotation already used by the renderers. It does not change the label rendering itself.

`tests/pcb/pcb-bounds-dimension-text-rotation.test.ts` contains six schema-validated regressions for explicit rotation, vertical dimensions and combined angles. All six fail against the unchanged base implementation and pass with the fix. The two new SVG snapshots contain the example above for both dimension types; existing snapshots are unchanged.

Validation: CI-pinned Bun 1.3.12 full suite, 387 pass / one existing TODO / zero failures; typecheck, build, format and dependency checks pass. Before/after renders were inspected, and the current PR CI and preview deployment pass. Text extents retain the existing approximate character-width/font-height model. AI-assisted implementation and review.


imrishabh18 (changes requested):
what's the motivation for this? reproduce a bug or motivation before making a fix

PR #758, opened 2026-09-18
Add PCB layer drawing order and opacity controls

KrishnaX12:
## Why

Multilayer PCB renderers need the same display controls PCB editors provide: configurable layer drawing order and layer transparency. Consumers currently have to rewrite colors to expose copper on lower layers, which couples view semantics to palette values.

## What changed

- add typed `layerDrawingOrder` and `layerOpacity` PCB SVG options
- define layer order from front to back while preserving default priorities for unlisted layers
- apply layer opacity as an SVG opacity multiplier without replacing colors or existing object opacity
- validate opacity values and reject duplicate drawing-order entries
- export the layer rendering types and document both options
- add focused coverage for color preservation, opacity, drawing order, and validation

## Verification

- `bun run format:check`
- `bunx tsc --noEmit`
- `bun run build`
- focused multilayer rendering tests: 5 passed
- full suite: 386 passed; one unrelated snapshot was flaky and passed immediately in isolation


KrishnaX12:
Closing at the requester’s direction; the fix will remain scoped to altium-to-circuit-json without a circuit-to-svg API change.

PR #760, opened 2026-09-19
fix(pcb): degrade non-finite silkscreen font_size and keepout radius to valid SVG lengths (#636)

### LINKED ISSUE #636: Non-finite silkscreen font_size and keepout radius emit font-size="NaN" and r="NaN"

marcos452652258-gif:
## Summary

Fixes tscircuit/circuit-to-svg#636.

When non-finite numbers (`NaN`) are present for silkscreen text `font_size` or circular keepout `radius`, the PCB SVG renderer previously emitted `font-size="NaN"` or `r="NaN"`, which are invalid SVG length values.

### Changes
- In `create-svg-objects-from-pcb-silkscreen-text.ts`, fallback `font_size` to the default `1` if non-finite or non-positive.
- In `create-svg-objects-from-pcb-keepout.ts`, fallback `radius` to `0` if non-finite or non-positive.
- Added unit test `tests/pcb/non-finite-silkscreen-keepout.test.ts`.


marcos452652258-gif:
Hi maintainers, pinging to keep this PR active — it is still tested, relevant and ready for review whenever you have time. Thank you!

PR #761, opened 2026-09-19
fix(sch): keep non-finite schematic component sizes out of the SVG (fixes #638)

### LINKED ISSUE #638: Non-finite schematic component size renders every coordinate as NaN, so the component vanishes

meanusarcanus:
## Summary

Fixes #638

When a `schematic_component` has non-finite dimensions (e.g. `NaN` or `Infinity` from an unparseable prop like `schWidth="abc"`), those non-finite values propagate through every derived coordinate calculation (`applyToPoint`, `componentScreenTopLeft`, `componentScreenBottomRight`, `componentScreenWidth`, `componentScreenHeight`). This emits `NaN` for `x`, `y`, `width`, `height`, and `transform` attributes across the component body rect, pin lines, and labels, causing SVG renderers to drop the elements and the component to vanish silently.

Additionally, non-finite sizes corrupt `getSchematicBoundsFromCircuitJson`, polluting the whole-sheet `viewBox` bounds.

### Fix
1. Add `getFiniteSchematicComponentSize` helper in `lib/sch/get-finite-schematic-component-size.ts` which treats non-finite `width` and `height` as `0`.
2. Apply `getFiniteSchematicComponentSize` in `getSchematicBoundsFromCircuitJson` when bounding `schematic_component` elements.
3. Apply `getFiniteSchematicComponentSize` in `createSvgObjectsFromSchematicComponentWithBox`.
4. Add comprehensive test cases in `tests/sch/non-finite-component-size.test.ts` verifying `NaN`, `Infinity`, and `-Infinity` dimensions produce clean, finite SVG without mutating input or altering finite component bounds.

@algora-pbc /claim #638


meanusarcanus:
### 💰 Payout Details
- **PayPal**: `tednadres@yahoo.com`

